PR #1, opened 2021-04-14
(#1949568) core: allow omitting second part of LoadCredentials= argument

lnykryn:
This allows "LoadCredentials=foo" to be used as shortcut for
"LoadCredentials=foo:foo", i.e. it's a very short way to inherit a
credential under its original name from the service manager into a
service.

(cherry picked from commit 8a29862e32fc7981fe16fc8dd79ad6863a8dc486)

Resolves: #1949568


dtardon (approved):
LGTM

PR #6, opened 2021-05-13
(#1931710) rfkill compilation error fixes

msekletar:

PR #7, opened 2021-05-14
(#1931710) rfkill: add some casts to silence -Werror=sign-compare

mrc0mmand:
Backport of the "correct" fix from upstream, since the fix from #6 wasn't complete and the build kept failing with:

```
In file included from ../src/basic/macro.h:491,
                 from ../src/basic/alloc-util.h:9,
                 from ../src/rfkill/rfkill.c:13:
../src/rfkill/rfkill.c: In function ‘run’:
../src/rfkill/rfkill.c:339:70: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 8 has type ‘int’ [-Werror=format=]
                         return log_error_errno(SYNTHETIC_ERRNO(EIO), "Short read of struct rfkill_event: (%zd < %lu)",
                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/basic/log.h:195:86: note: in definition of macro ‘log_full_errno’
                         ? log_internal(_level, _e, PROJECT_FILE, __LINE__, __func__, __VA_ARGS__) \
                                                                                      ^~~~~~~~~~~
../src/rfkill/rfkill.c:339:32: note: in expansion of macro ‘log_error_errno’
                         return log_error_errno(SYNTHETIC_ERRNO(EIO), "Short read of struct rfkill_event: (%zd < %lu)",
                                ^~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

/cc @dtardon @msekletar

### DIFF
--- a/src/rfkill/rfkill.c
+++ b/src/rfkill/rfkill.c
@@ -177,7 +177,7 @@ static int load_state(Context *c, const struct rfkill_event *event) {
         ssize_t l = write(c->rfkill_fd, &we, sizeof we);
         if (l < 0)
                 return log_error_errno(errno, "Failed to restore rfkill state for %i: %m", event->idx);
-        if (l < RFKILL_EVENT_SIZE_V1)
+        if ((size_t)l < RFKILL_EVENT_SIZE_V1) /* l cannot be < 0 here. Cast to fix -Werror=sign-compare */
                 return log_error_errno(SYNTHETIC_ERRNO(EIO),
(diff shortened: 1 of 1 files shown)


msekletar:
Ahh, I was looking for upstream fix but I didn't realise that I had checked out older tree that didn't contain the fix and I falsely concluded this wasn't fixed upstream yet. My assumption was that maybe this is caused by compiler version difference between Fedora (upstream) and RHEL-9.

msekletar (approved):
LGTM

PR #8, opened 2021-07-01
(#1976832) rules: load sg module

dtardon:
(cherry picked from commit a42b57dc8b265f183a8fb6fe9ae32a9d77cbb7c5)
(cherry picked from commit fab2dff96f59e0851884b4ef32dccab763f5eef1)

RHEL-only

Resolves: #1976832


mrc0mmand:
The CI fails should be fixed by backporting https://github.com/systemd/systemd/commit/7c5fd25119a495009ea62f79e5daec34cc464628 - not sure if we want to, though, since we're going to rebase RHEL9 systemd anyway.

dtardon:
This doesn't really need CI anyway, does it?

mrc0mmand:
> This doesn't really need CI anyway, does it?

Yeah, I don't think it does.

PR #14, opened 2021-07-21
(#1984299) basic/unit-name: do not use strdupa() on a path

jamacku:
The path may have unbounded length, for example through a fuse mount.

CVE-2021-33910: attacked controlled alloca() leads to crash in systemd and
ultimately a kernel panic. Systemd parses the content of /proc/self/mountinfo
and each mountpoint is passed to mount_setup_unit(), which calls
unit_name_path_escape() underneath. A local attacker who is able to mount a
filesystem with a very long path can crash systemd and the whole system.
    
https://bugzilla.redhat.com/show_bug.cgi?id=1970887
    
The resulting string length is bounded by UNIT_NAME_MAX, which is 256. But we
can't easily check the length after simplification before doing the
simplification, which in turns uses a copy of the string we can write to.
So we can't reject paths that are too long before doing the duplication.
Hence the most obvious solution is to switch back to strdup(), as before
7410616cd9dbbec97cf98d75324da5cda2b2f7a2.
    
Resolves: #1984299
    
(cherry picked from commit 441e0115646d54f080e5c3bb0ba477c892861ab9)

### DIFF
--- a/src/basic/unit-name.c
+++ b/src/basic/unit-name.c
@@ -528,7 +528,7 @@ int unit_name_from_path(const char *path, const char *suffix, char **ret) {
         if (strlen(s) >= UNIT_NAME_MAX) /* Return a slightly more descriptive error for this specific condition */
                 return -ENAMETOOLONG;
 
-        /* Refuse this if this got too long or for some other reason didn't result in a valid name */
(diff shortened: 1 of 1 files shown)


msekletar (approved):
LGTM

PR #15, opened 2021-07-27
Check return value of pam_get_item/pam_get_data functions (#1973210)

jacekmigacz:
Already merged upstream: https://github.com/systemd/systemd/pull/19965/


dtardon (approved):
LGTM

PR #18, opened 2021-08-25
(#1997200) pid1: bump DefaultTasksMax to 80% of the kernel pid.max value

dtardon:
This should be hopefully high enough even for the very big deployments.

Resolves: #1997200

### DIFF
--- a/man/systemd-system.conf.xml
+++ b/man/systemd-system.conf.xml
@@ -389,10 +389,10 @@
         <listitem><para>Configure the default value for the per-unit <varname>TasksMax=</varname> setting. See
         <citerefentry><refentrytitle>systemd.resource-control</refentrytitle><manvolnum>5</manvolnum></citerefentry>
         for details. This setting applies to all unit types that support resource control settings, with the exception
-        of slice units. Defaults to 15% of the minimum of <varname>kernel.pid_max=</varname>, <varname>kernel.threads-max=</varname>
+        of slice units. Defaults to 80% of the minimum of <varname>kernel.pid_max=</varname>, <varname>kernel.threads-max=</varname>
         and root cgroup <varname>pids.max</varname>.
         Kernel has a default value for <varname>kernel.pid_max=</varname> and an algorithm of counting in case of more than 32 cores.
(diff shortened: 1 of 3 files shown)


man/systemd-system.conf.xml
```diff
@@ -389,10 +389,10 @@
         <listitem><para>Configure the default value for the per-unit <varname>TasksMax=</varname> setting. See
         <citerefentry><refentrytitle>systemd.resource-control</refentrytitle><manvolnum>5</manvolnum></citerefentry>
         for details. This setting applies to all unit types that support resource control settings, with the exception
-        of slice units. Defaults to 15% of the minimum of <varname>kernel.pid_max=</varname>, <varname>kernel.threads-max=</varname>
+        of slice units. Defaults to 80% of the minimum of <varname>kernel.pid_max=</varname>, <varname>kernel.threads-max=</varname>
         and root cgroup <varname>pids.max</varname>.
         Kernel has a default value for <varname>kernel.pid_max=</varname> and an algorithm of counting in case of more than 32 cores.
-        For example with the default <varname>kernel.pid_max=</varname>, <varname>DefaultTasksMax=</varname> defaults to 4915,
+        For example with the default <varname>kernel.pid_max=</varname>, <varname>DefaultTasksMax=</varname> defaults to 26214,
```
msekletar, line 395: `kernel.pid_max=` on RHEL-9 is set to 4194304 and 80% of that is 3355443.

msekletar (changes requested):

man/systemd-system.conf.xml
dtardon: Hm, right. That makes 15% 629145, which should be enough, I think? So we don't need to change this at all.

msekletar: Make sense. Let's try to go with default then and if anyone complains (I doubt) we will bump it up.

dtardon: I was too fast... The 100% value is min(kernel.pid_max, kernel.threads-max) = 11403. So we do need this.

jamacku (approved):
LGTM

PR #20, opened 2021-08-26
(#1998190) rules: add elevator= kernel command line parameter

dtardon:
Kernel removed the elevator= option, so let's reintroduce
it for rhel8 via udev rule.

RHEL-only

Resolves: #1998190


msekletar (approved):
LGTM

PR #21, opened 2021-08-30
set core ulimit to 0 like on RHEL-7

dtardon:
RHEL-only

Resolves: #198509


mrc0mmand (approved):
LGTM

msekletar (changes requested):
Commit message references incorrect bug. The bug number should be instead 1998509

dtardon:
> Commit message references incorrect bug. The bug number should be instead 1998509

Fixed.

msekletar (approved):
LGTM

PR #24, opened 2021-09-20
(#1972223) boot: don't build bootctl when -Dgnu-efi=false is set

msekletar:


msekletar:
Note that there are specfile changes that need be introduced to fully fix the issue reported in https://bugzilla.redhat.com/show_bug.cgi?id=1972223

msekletar:
We need to have `-Defi=true -Dgnu-efi=false` in `CONFIGURE_OPTS=`

dtardon (approved):
LGTM

PR #27, opened 2021-09-21
pid1: bump DefaultTasksMax to 80% of the kernel pid.max value

dtardon:
This should be hopefully high enough even for the very big deployments.

RHEL-only

Resolves: #2003031


jamacku (approved):
LGTM

PR #29, opened 2021-09-21
(#1921094) udev/net-setup-link: change the default MACAddressPolicy to "none"

msekletar:
While stable MAC address for interface types that don't have the
address provided by HW could be useful it also breaks LACP based bonds.
Let's err on the side of caution and don't change the MAC address from
udev.

Resolves: #1921094


dtardon (approved):
LGTM

PR #31, opened 2021-09-22
(#1972223)  unit: install the systemd-bless-boot.service only if we have gnu-efi

msekletar:
Follow-up to #20591.

(cherry picked from commit 220261ef940a126588b20a1765a2501811473839)

Related: #1972223


dtardon (approved):
LGTM

PR #33, opened 2021-09-22
(#2000927) units: don't enable tmp.mount statically in local-fs.target

msekletar:
Related: #2000927


dtardon:
The 9.0.0 bug is https://bugzilla.redhat.com/show_bug.cgi?id=2000927 . I'm not sure the dist-git check will allow this to be pushed with the 9.0-beta bug reference...

dtardon:
> The 9.0.0 bug is https://bugzilla.redhat.com/show_bug.cgi?id=2000927 . I'm not sure the dist-git check will allow this to be pushed with the 9.0-beta bug reference...

Looking at gitbz code, it looks like it will, as gitbz goes through Blocks list of the referenced bug if the bug itself doesn't have correct release.

dtardon (approved):
LGTM

msekletar:
I will open new PR and fix the bug reference in commit message.

msekletar:
Closed in favor of #35.

PR #38, opened 2021-10-01
(#1921094) udev/net-setup-link: *really* change the default MACAddressPolicy to "none"

msekletar:
Fix the oversight and change the policy in the link file, i.e. the
place where it actually matters.

Related: #1921094


dtardon (approved):
LGTM

PR #42, opened 2021-10-21
(#2016042) openssl-util: use EVP API to get RSA bits

dtardon:


dtardon:
@mrc0mmand I've added https://github.com/systemd/systemd/pull/21073 here so I don't have to create a separate bug for it...

mrc0mmand:
> @mrc0mmand I've added [systemd/systemd#21073](https://github.com/systemd/systemd/pull/21073) here so I don't have to create a separate bug for it...

Excellent, thanks! With this merged, I can finally switch CentOS CI to the C9S nodes.

PR #47, opened 2021-10-27
(#2005367) (#2005367) Drop NAMING_REPLACE_STRICTLY from naming scheme

jacekmigacz:
It allows UTF-8 character in interface name.


msekletar (changes requested):
Please adjust man pages the same way it is done on RHEL-8

jacekmigacz:
Closing to be reopened.

PR #48, opened 2021-11-02
(#2019468) core/mount: add implicit unit dependencies even if when mount unit is…

dtardon:
… generated from /proc/self/mountinfo

Hopefully fixes #20566.

(cherry picked from commit aebff2e7ce209fc2d75b894a3ae8b80f6f36ec11)

Resolves: #2019468


jacekmigacz (approved):
LGTM

PR #50, opened 2021-11-04
(#2017035) ci: use C9S chroots in Packit

mrc0mmand:
rhel-only
Related: #2017035

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -37,9 +37,8 @@ jobs:
   trigger: pull_request
   metadata:
     targets:
-      # FIXME: change to CentOS 9 once it's available
-      - fedora-34-x86_64
-      - fedora-34-aarch64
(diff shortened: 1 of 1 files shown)


mrc0mmand:
The Packit jobs would benefit greatly if we reintroduced `%check` in our specfile (basically reverting https://gitlab.com/redhat/centos-stream/rpms/systemd/-/commit/6545fb6a47cb864f71d291f65e088630bcf4cb6c#6243b7b4200e5ce63aec17014a3882cb22d6c2ee_617_622). If that would be PITA, I can *sed* it back (basically adding a yet another `sed` to the already growing collection), but I'd prefer the former (and cleaner) solution.

/cc @jamacku @dtardon @msekletar

jamacku (approved):
LGTM

PR #57, opened 2022-01-10
(#2006762) resolved: synthesize option

jacekmigacz:
It is currently possible to query with `resolvectl` and provide `--synthesize=no` flag for non-synthesized answers.

This patch provides systemd-resolved synthesize control switch and makes possible to disable synthesized answers for the regular DNS questions (like these emitted from `nslookup` or `dig`).


jacekmigacz:
Moved to https://github.com/redhat-plumbers/systemd-rhel9/pull/64

PR #64, opened 2022-02-01
(#2006762) resolved: synthesize option

jacekmigacz:
It is currently possible to query with `resolvectl` and provide `--synthesize=no` flag for non-synthesized answers.

This patch provides systemd-resolved synthesize control switch and makes possible to disable synthesized answers for the regular DNS questions (like these emitted from `nslookup` or `dig`).


msekletar:
@jacekmigacz You need to prepare upstream PR and get it merged. We can't merge this downstream before this lands in systemd/systemd. Thanks!

jacekmigacz:
This is a lightweight version of the original patch; limited to controlling hostname synthesis.

jamacku:

PR #66, opened 2022-02-02
(#1982596) man: mention System Administrator's Guide in systemctl manpage

dtardon:
RHEL-only

Resolves: #1982596


dtardon:
Note: only 9-beta docs exist ATM, but I hope the URL for 9 will be the same (except for 9-beta -> 9, of course)...

jamacku (changes requested):
Listed page doesn't work. 

Update: Oh, I see, what you meant by your comment. But still, I would wait before merging this.

jamacku (approved):
LGTM, let's put it into 9.0.0

PR #67, opened 2022-02-03
(#2017035) tests: rework test macros to not take code as parameters

dtardon:
I suspect we'd end up backporting this sooner or later anyway, so let's do it now while it's easy...


mrc0mmand (approved):
LGTM, thanks! This will definitely help in the foreseeable future.

dtardon:
Fixed the bug number...

mrc0mmand (approved):
Still LGTM.

PR #68, opened 2022-02-04
(#2052106) Net naming scheme for RHEL-9.0

jacekmigacz:


msekletar:
@jacekmigacz thank you preparing the PR! However, I am now looking at the associated BZ and to be quite honest I don't see a good reason do deviate from upstream naming scheme. I will request more information from the reporter.

msekletar:
I was about to LGTM on this one, but I've noticed that in the original BZ reported said it is not worth to deviate from the upstream naming scheme, so I've closed that bug.

@jacekmigacz I've created new bug requesting the addition of `rhel-9.0` naming scheme (w/o changes that were originally requested) and I've also updated description of this PR. Please update `Resolves:` in your commit message and update this PR. Thanks!

jacekmigacz:
@msekletar thanks for creating bz for me I forgot to do it. Force-pushed with new bz# in `Resolves:`.

msekletar (approved):
LGTM

PR #72, opened 2022-03-08
(#2013213) (Not only) CI and QoL backports

mrc0mmand:
So, uh, this turned out to be bigger than I expected, mainly due to the meson-related changes. I could rewrite the patches and thus avoid the need to backport them, but this should make backporting easier in the long run.


jamacku:
wow

dtardon (approved):
LGTM

jamacku (commented):
@mrc0mmand - referenced bug [#2013213](https://bugzilla.redhat.com/show_bug.cgi?id=2013213) is for RHEL 9.0.0 and PR is targeted at main, is it correct?

mrc0mmand:
> @mrc0mmand - referenced bug [#2013213](https://bugzilla.redhat.com/show_bug.cgi?id=2013213) is for RHEL 9.0.0 and PR is targeted at main, is it correct?

Yup. The original goal was to have one CI-related tracker for each major RHEL release (at least that's what @dtardon envisioned). Hopefully it still holds.

jamacku:
> Yup. The original goal was to have one CI-related tracker for each major RHEL release (at least that's what @dtardon envisioned). Hopefully it still holds.

@mrc0mmand But I'm not sure if we would be able to pass checks on C9S MR.

dtardon:
> > @mrc0mmand - referenced bug [#2013213](https://bugzilla.redhat.com/show_bug.cgi?id=2013213) is for RHEL 9.0.0 and PR is targeted at main, is it correct?
> 
> Yup. The original goal was to have one CI-related tracker for each major RHEL release (at least that's what @dtardon envisioned). Hopefully it still holds.

It does, but the problem here is that the referenced bug *is not* the RHEL-9 CI tracker...  (Yeah, it's partly my fault too, as I should have checked it.)

mrc0mmand:
> > > @mrc0mmand - referenced bug [#2013213](https://bugzilla.redhat.com/show_bug.cgi?id=2013213) is for RHEL 9.0.0 and PR is targeted at main, is it correct?
> > 
> > 
> > Yup. The original goal was to have one CI-related tracker for each major RHEL release (at least that's what @dtardon envisioned). Hopefully it still holds.
> 
> It does, but the problem here is that the referenced bug _is not_ the RHEL-9 CI tracker... (Yeah, it's partly my fault too, as I should have checked it.)

:facepalm: I have no idea where I came up with that bug number then, *sigh*

PR #79, opened 2022-04-06
(#2070097) udev/net_id: avoid slot based names only for single function devices

msekletar:
If we have two or more devices that share the same slot but they are
also multifunction then it is OK to use the slot information even if it
is the same for all of them. Name conflict will be avoided because we
will append function number and form names like, ens1f1, ens1f2...

(cherry picked from commit 66425daf2c68793adf24a48a26d58add8662e83f)

Resolves: #2070097

### DIFF
--- a/man/systemd.net-naming-scheme.xml
+++ b/man/systemd.net-naming-scheme.xml
@@ -406,7 +406,12 @@
         <varlistentry>
           <term><constant>rhel-9.0</constant></term>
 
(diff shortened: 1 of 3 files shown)


dtardon (approved):
LGTM

jamacku (commented):
Looks like the CentOS CI fail is unrelated and will be fixed in 9.1 by https://github.com/redhat-plumbers/systemd-rhel9/pull/72.

\cc @mrc0mmand

PR #80, opened 2022-04-16
(#2075987) Backport udev/hwdb support for AV production controllers

Conan-Kudo:
This backports https://github.com/systemd/systemd/pull/22730 to systemd 250 to allow [Boatswain](https://feaneron.com/2022/03/17/boatswain-your-stream-deck-app-for-linux/) to work properly on RHEL/CentOS 9.

cc: @GeorgesStavracas, @cschalle

Resolves: [rhbz#2075987](https://bugzilla.redhat.com/show_bug.cgi?id=2075987)

### DIFF
--- a/hwdb.d/70-av-production.hwdb
+++ b/hwdb.d/70-av-production.hwdb
@@ -1,0 +1,123 @@
+# This file is part of systemd.
+#
+# Database for AV production controllers that should be accessible to the seat owner.
+#
+# This covers DJ tables, and music-oriented key pads
+#
+# To add local entries, copy this file to
+#   /etc/udev/hwdb.d/
+# and add your rules there. To load the new rules execute (as root):
+#   systemd-hwdb update
+#   udevadm trigger
+
+################
+# Ableton
+################
+# Push 2
+usb:v2982p1967*
+ ID_AV_PRODUCTION_CONTROLLER=1
+
+################
+# Eks
+################
+# Otus
+usb:v1157p0300*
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
cc: @davide125, @davdunc, @dbrandonjohnson, @michel-slm

GeorgesStavracas (approved):

jamacku (commented):
Thank you for the proposed PR, but since these changes aren't present in Fedora yet. I would wait before introducing these changes to RHEL9.

jamacku:
> I believe @keszybz will be shipping it to Fedora when updating to systemd 250.5 (which includes these fixes): https://github.com/systemd/systemd-stable/releases/tag/v250.5, so I would consider this not a blocker from that perspective.

Based on https://src.fedoraproject.org/rpms/systemd/ It's currently present only in Rawhide in `systemd-251~rc2-1.fc37`

keszybz:
The patch is also in v249.12, which is in `updates-testing` for F35. Unrelated issues are blocking that update from graduating to stable, but it'll get there at some point.

jamacku:
Thank you @keszybz, so I would wait at least till `v249.12` is stable.

Conan-Kudo:
> Thank you @keszybz, so I would wait at least till `v249.12` is stable.

This has been done for a while now. Can we get this merged and released into RHEL 9?

dtardon:
We plan to do regular HWDB updates starting with 9.1. I don't think there's any point in backporting  a few commits separately.

PR #82, opened 2022-05-10
(#2047682) shutdown: get only active md arrays.

dtardon:
Current md_list_get() implementation filters all block devices, started from
"md*". This is ambiguous because list could contain:
- partitions created upon md device (mdXpY)
- external metadata container- specific type of md array.

For partitions there is no issue, because they aren't handle STOP_ARRAY
ioctl sent later. It generates misleading errors only.

Second case is more problematic because containers are not locked in kernel.
They are stopped even if container member array is active. For that reason
reboot or shutdown flow could be blocked because metadata manager cannot be
restarted after switch root on shutdown.

Add filters to remove partitions and containers from md_list. Partitions
can be excluded by DEVTYPE. Containers are determined by MD_LEVEL
property, we are excluding all with "container" value.

Signed-off-by: Mariusz Tkaczyk <mariusz.tkaczyk@linux.intel.com>
(cherry picked from commit 3a3b022d2cc112803ea7b9beea98bbcad110368a)

### DIFF
--- a/src/shutdown/umount.c
+++ b/src/shutdown/umount.c
@@ -352,14 +352,30 @@ static int md_list_get(MountPoint **head) {
         if (r < 0)
                 return r;
 
+        /* Filter out partitions. */
+        r = sd_device_enumerator_add_match_property(e, "DEVTYPE", "disk");
+        if (r < 0)
+                return r;
+
         FOREACH_DEVICE(e, d) {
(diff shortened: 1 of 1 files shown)


jamacku (commented):
@dtardon There is missing Bugzilla reference.

msekletar (approved):
LGTM

PR #83, opened 2022-05-18
(#2087677) shutdown: get only active md arrays.

dtardon:
Current md_list_get() implementation filters all block devices, started from
"md*". This is ambiguous because list could contain:
- partitions created upon md device (mdXpY)
- external metadata container- specific type of md array.

For partitions there is no issue, because they aren't handle STOP_ARRAY
ioctl sent later. It generates misleading errors only.

Second case is more problematic because containers are not locked in kernel.
They are stopped even if container member array is active. For that reason
reboot or shutdown flow could be blocked because metadata manager cannot be
restarted after switch root on shutdown.

Add filters to remove partitions and containers from md_list. Partitions
can be excluded by DEVTYPE. Containers are determined by MD_LEVEL
property, we are excluding all with "container" value.

Signed-off-by: Mariusz Tkaczyk <mariusz.tkaczyk@linux.intel.com>
(cherry picked from commit 3a3b022d2cc112803ea7b9beea98bbcad110368a)

Resolves: #2087677

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -28,8 +28,14 @@ jobs:
       - name: Repository checkout
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
       - name: Install build dependencies
-        run: sudo -E .github/workflows/unit_tests.sh SETUP
+        run: |
+          # Drop XDG_* stuff from /etc/environment, so we don't get the user
+          # XDG_* variables when running under sudo
+          sudo sed -i '/^XDG_/d' /etc/environment
+          # Pass only specific env variables through sudo, to avoid having
+          # the already existing XDG_* stuff on the "other side"
+          sudo --preserve-env=CRYPTOLIB,GITHUB_ACTIONS,CI .github/workflows/unit_tests.sh SETUP
       - name: Build & test (${{ matrix.run_phase }}-${{ matrix.cryptolib }})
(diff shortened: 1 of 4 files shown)


jamacku:
@mrc0mmand Could you please have a look at CentOS CI

mrc0mmand:
Ahh, this PR is open against 9.0.0, which is missing https://github.com/redhat-plumbers/systemd-rhel9/commit/fcaf368341c85285017f77333543656e1382975e. Backporting it should make CI happy again.

jamacku:
> Ahh, this PR is open against 9.0.0, which is missing [fcaf368](https://github.com/redhat-plumbers/systemd-rhel9/commit/fcaf368341c85285017f77333543656e1382975e). Backporting it should make CI happy again.

Thank you, @dtardon could you please backport [fcaf368](https://github.com/redhat-plumbers/systemd-rhel9/commit/fcaf368341c85285017f77333543656e1382975e), Thanks

mrc0mmand:
The TEST-58-REPART fail is unrelated, it should be fixed by one of the backport PRs (I guess https://github.com/redhat-plumbers/systemd-rhel9/pull/72 is the one).

mrc0mmand:
Aand the memleak can be fixed by one of the patches from #86 - [2aaf6d407e8541985a15b7106abf6fbdfed0766a](https://github.com/redhat-plumbers/systemd-rhel9/pull/86/commits/6a5ad6b650ad8ca23d6eab897da920f9686fa6ce).

msekletar (approved):
LGTM

PR #84, opened 2022-05-30
(#2087652) ci(Mergify): Add rules for `needs-ci` label management

jamacku:
Add rules for `needs-ci` label management

RHEL-only

Related: #2087652

---

_This change has been made by @jamacku from https://mergify.com config editor._


jamacku:
There should be nothing that blocks this PR, merging.

PR #85, opened 2022-06-06
(#2087652) ci(Mergify): fix copy&paste bug

dtardon:
RHEL-only

Related: #2087652


jamacku (approved):
Great catch! Thank you David!

PR #86, opened 2022-06-13
Several assorted patches for memory/fd leaks

mrc0mmand:
So I can extend the list of tests we run under sanitizers for RHEL 9 (see https://github.com/systemd/systemd-centos-ci/pull/502).


dtardon (changes requested):

dtardon (approved):
It's discouraging to see how many resource leaks sneak through review...

PR #87, opened 2022-06-21
(#2082131) Fixups to the unit enablement logic

dtardon:
All right, this is a big one... I've "hijacked" the bug to also backport https://github.com/systemd/systemd/pull/22649, which is vaguely related. The PR fixes various corner cases for `enable` and `link` (and adds tests for all this). In addition to these fixes, there are also several tree-wide changes there that will likely make future backports a bit easier.

If this is not wanted--or considered too risky--I can publish just the backport @mrc0mmand wanted; that's just 3 commits...

Resolves: [#2082131](https://bugzilla.redhat.com/show_bug.cgi?id=2082131)


mrc0mmand:
@dtardon if you include https://github.com/redhat-plumbers/systemd-rhel9/pull/89/commits/51aa12c8027215117f78ade6ed6673d548d45cd7 (from #89), the CIs should be happy.

Anyway, I'd definitely go with this PR, as it makes backporting stuff much easier (see #89, which is already rebased on top of this one, exactly for that reason).

dtardon:
> @dtardon if you include [51aa12c](https://github.com/redhat-plumbers/systemd-rhel9/commit/51aa12c8027215117f78ade6ed6673d548d45cd7) (from #89), the CIs should be happy.

Thanks, I've done that already; I just haven't pushed it yet.

dtardon:
> > @dtardon if you include [51aa12c](https://github.com/redhat-plumbers/systemd-rhel9/commit/51aa12c8027215117f78ade6ed6673d548d45cd7) (from #89), the CIs should be happy.
> 
> Thanks, I've done that already; I just haven't pushed it yet.

(I didn't see the build failure locally because I usually build using RHEL configuration.)

msekletar (approved):
LGTM

I did very basic sanity check of the backport and it seems OK, but it is huge so I may have missed something. At any rate, it is early in RHEL-9 lifecycle and it will make future backports easier so let's try our luck with this one.

PR #88, opened 2022-06-22
(#2087652) Backport TEST-21-DFUZZER

mrc0mmand:


test/units/testsuite-21.sh
```diff
@@ -27,18 +27,18 @@ systemctl log-level info
 # TODO
 #   * check for possibly newly introduced buses?
 BUS_LIST=(
-    org.freedesktop.home1
+#    org.freedesktop.home1
```
dtardon, line 30: Perhaps the list could be filtered for names that can be activated? Something like
```sh
BUS_LIST=( $(busctl list --no-legend --activatable | cut -d ' ' -f 1 | sort | join - <<<"$BUS_LIST[@]") )
```
(untested).

It would be even simpler to make the file a template and make inclusion of the names conditional, but I've no idea if the test infrastructure does look for files in builddir too...

In any case, that's an idea for upstream improvement. This is good enough for us.

dtardon (approved):
LGTM

PR #89, opened 2022-06-22
Backport "a couple" of (mostly) test-related commits

mrc0mmand:
@dtardon I see your #87 and raise a worthy competitor!

This, mainly, backports TEST-35-LOGIN, TEST-45-TIMEDATE, TEST-70-TPM2, and TEST-73-LOCALE, which were recently introduced in upstream.

Also, note: this is rebased on top of #87 (since it requires several of the patches already backported there), hence the enormous commit count.


mrc0mmand:
I dropped the new TEST-64 stuff, since it depends on many patches for udevadm and similar, which would make this PR unreviewable (it's already on a thin ice).

dtardon (approved):
LGTM

PR #94, opened 2022-07-19
(#2073003) udev/net_id: avoid slot based names only for single function devices

msekletar:
If we have two or more devices that share the same slot but they are
also multifunction then it is OK to use the slot information even if it
is the same for all of them. Name conflict will be avoided because we
will append function number and form names like, ens1f1, ens1f2...

(cherry picked from commit 66425daf2c68793adf24a48a26d58add8662e83f)

Resolves: #2073003


dtardon (approved):
LGTM

PR #100, opened 2022-08-17
Revert "shared/install: create relative symlinks for enablement and a…

lnykryn:
…liasing"

This reverts commit d6c9411072901556176ac130f2ce71a33107aa93.

I still think this is something that needs to be done, but we're hitting some
unexplained failures, e.g. https://github.com/systemd/systemd/issues/22920.
So let's revert this for now, so -rc2 can be released, with a plan to return
to this after a release.

Closes #22920.

(cherry picked from commit 9aa3d6b41eadff6634bb3cf800064724171a7aad)

Resolves: #2118668

### DIFF
--- a/src/shared/install.c
+++ b/src/shared/install.c
@@ -1904,7 +1904,7 @@ static int install_info_symlink_alias(
                 if (!alias_path)
                         return -ENOMEM;
 
-                q = create_symlink(lp, info->name, alias_path, force, changes, n_changes);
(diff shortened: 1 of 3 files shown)


mrc0mmand:
Oh god, I didn't notice we pulled this in as well :facepalm:.

mrc0mmand (approved):

jamacku (approved):

PR #101, opened 2022-08-22
shared/install: fix crash when reenable is called without --root

lnykryn:
(cherry picked from commit 6e961aeb262521742a4cd92e4620de193f159f7c)

Resolves: #2120222


jamacku (approved):
LGTM

PR #105, opened 2022-09-02
(#2100464) Add option to stop idle sessions after specified timeout

msekletar:


test/units/testsuite-35.sh
```diff
@@ -388,6 +390,23 @@ test_session_properties() {
...
+    if [[ ! -c /dev/tty2 ]]; then
+        echo "/dev/tty2 does not exist, skipping test ${FUNCNAME[0]}."
+        return
+    fi
+
+    trap cleanup_session RETURN
+    create_session
+
+    assert_eq "$(loginctl list-users --no-legend | awk '$2 == "logind-test-user" { print $1 }')" "$(id -ru logind-test-user)"
+    assert_eq "$(loginctl list-users --no-legend | awk '$2 == "logind-test-user" { print $3 }')" no
```
dtardon, line 432: The asserts on lines 403 and 407 will fail, because the 3rd column showing linger status has only been added recently.

msekletar: Frankly, I didn't expect CI to PASS on the first try. Because I didn't test it locally before pushing this PR draft.

I've now changed these two problematic lines like so,
```bash
assert_eq "$(loginctl  show-user --property Linger --value logind-test-user)" no
assert_eq "$(loginctl  show-user --property Linger --value logind-test-user)" yes
```

msekletar:
@mrc0mmand do you have any idea about CentOS Stream 9 CI failure?

mrc0mmand:
> @mrc0mmand do you have any idea about CentOS Stream 9 CI failure?

That's unrelated to this PR, we're just missing a lot of recent dissect/lodev related patches.

dtardon (approved):
LGTM

jamacku:
Commits in this PR reference RHEL 9.1.0 bug.

PR #106, opened 2022-09-07
(#2122500) (#1985288) rules: import previous SYSTEMD_READY state for suspended DM devices

msekletar:
…nd skip other rules

We can't get any FS meta-data from a suspended device. Hence defer making any plugged/unplugged decisions, i.e. we just import whatever was previous state and skip processing all other rules.

Thanks Lennart Poettering <lennart@poettering.net> for suggesting this solution.

(cherry picked from commit 466266c1728e3b53b7020e4c0165c3e88576b805)

Resolves: #1985288


jamacku:

PR #107, opened 2022-09-15
(#2122500) Use CodeQL instead of LGTM

mrc0mmand:
as LGTM is going to be shut down by EOY. Also, pull in a couple of other CodeQL-related tweaks.


mrc0mmand:
The LGTM fail can be safely ignored (I disabled LGTM completely after opening this PR, so it won't show up again). Also, the CodeQL fail is also a one time thing (unless someone introduces a new alert in a future PR), as I renamed the action file, so it's missing base results to compare against.

dtardon (approved):
LGTM

.github/workflows/codeql.yml
```diff
@@ -6,16 +6,19 @@ name: "CodeQL"
 
 on:
   pull_request:
-    branches: [main]
+    branches:
+      - main
+      - v[0-9]+-stable
```
jamacku, line 11: So far looks good. Just don't forget to change branches to match our source-git branches.

jamacku (approved):

PR #110, opened 2022-09-22
(#2100464) logind: schedule idle check full interval from now if we couldn't figure out atime timestamp

msekletar:
(cherry picked from commit 6edf707fd59347024fa6be0342b108527825db1f)

Related: #2100464


msekletar:
@jamacku this contains bunch of other commits that are in `main` but aren't in `rhel-9.1`. I think we need to merge `main` into`rhel-9.1` and then this PR will contain only single commit.

msekletar:
Also I didn't opened this against `main` because for that branch we will need a different PR with different BZ number.

jamacku:
> @jamacku this contains bunch of other commits that are in `main` but aren't in `rhel-9.1`. I think we need to merge `main` into`rhel-9.1` and then this PR will contain only single commit.

@msekletar We can't merge `main` on top of `rhel-9.1`, since it already has some `rhel-9.2` commits. And also, I have just noticed that #105 was merged into the wrong branch.

We are at the point when we should follow the Z-stream process for RHEL 9.1 bugs. I'm going to try to fix the `main` branch and create a new PR as a replacement for #105.

jamacku:
I have created a new PR - #111 with the commit ` logind: schedule idle check full interval from now if we couldn't figure out a time timestamp`.

@msekletar, please have a look. I hope I have cherry-picked the correct commit from upstream since you have probably referenced the commit from your branch.

I'm closing this PR since it contains commits we can't have in `rhel-9.1`.

Issue #96, opened 2022-07-19
Backport `udevadm wait` & `udevadm lock` stuff


mrc0mmand:
It would be great to have this in RHEL 9 (especially for making tests more reliable, like TEST-64-UDEV-STORAGE and TEST-58-REPART), however, it might pull in a lot of other stuff.

Main PRs:
 - https://github.com/systemd/systemd/pull/22872
 - https://github.com/systemd/systemd/pull/22867

Follow-ups:
 - https://github.com/systemd/systemd/pull/22960 (depends on https://github.com/systemd/systemd/pull/22745)
 - https://github.com/systemd/systemd/pull/22964
 - https://github.com/systemd/systemd/pull/22996
 - and more (probably)

jamacku:
Done via rebase to v252.

PR #108, opened 2022-09-15
(#2122500) A couple of test-related backports

mrc0mmand:
~~Unfortunately, this is missing a lot of TEST-64/TEST-50/etc. related commits, which are blocked by #96, making the tests unreliable and lacking the latest code coverage/fixes.~~

Currently built on top of #113, so the commit count might be *slightly* higher.

PR #113, opened 2022-10-13
(#2087652) Backport udevadm wait & udevadm lock stuff

dtardon:
Fixes: #96


mrc0mmand:
Missing patches (in order to be simply cherry-picked on top of this PR):

```
928c6c00c37cc01860a0a6fec30e8188bb52929b
c412f4c603182ad16636d1747c952c12f18840c6
39fc0174c39a32502ea215d1676491f6ce46e2e7
224c853fdf1c0dfb918b1f9eaa79987dda2d9b60
695cfd53fedcdb4f1ee5d5a7e6f2f4870c0e1fd7
400e3d21f8cae53a8ba9f9567f244fbf6f3e076c
4565246911adbdd1b20d8944c0754772788a768c
655eb07391aef911d7f2246e9fd1f0ee29de6a03
5996f1167931d9b5760c368665104fc23f47de0f
```

This all leads to a commit which merges test-repart.sh and TEST-58-REPART together - 2b21522e82f920a8541df4daf0b65fdded277276 - but there's a small conflict, since the backported https://github.com/redhat-plumbers/systemd-rhel9/pull/113/commits/859bf63e1a0694bb83ef694765827eccae66fbdc is missing the original changes in the meson.build files, but it should be pretty easy to fix.

Then there's the last commit from *that* PR (systemd/systemd#24373) - 760971125dc8c3241694db90a2b5357f812cd9dd - which should, hopefully, apply cleanly after the conflict is resolved.

mrc0mmand:
So, I rebased #108 on top of this PR, and with it TEST-64 finally passes in both CentOS CI runs. However, there's another issues with TEST-50-DISSECT and TEST-58-REPART, which is (iirc) related to https://github.com/systemd/systemd/pull/23427 and https://github.com/systemd/systemd/pull/24530 (but I'm not completely sure), but my attempt in backporting this led to another rabbit hole, so it might be... a bit involved.

jamacku:
@dtardon I think this is no longer needed since we have rebased to v252.

dtardon:
Yes, of course it's not.

PR #114, opened 2022-10-18
(#2031810) Support by-path devlink for multipath NVMe block devices

dtardon:
Resolves: [#2031810](https://bugzilla.redhat.com/show_bug.cgi?id=2031810)


jamacku:

PR #118, opened 2022-11-29
(#2149074) Avoid deadlock when passing processed backtrace data

dtardon:
Resolves: [#2149074](https://bugzilla.redhat.com/show_bug.cgi?id=2149074)


msekletar (approved):
LGTM

PR #122, opened 2022-12-05
(#2141979) test: check if we can use SHA1 MD for signing before using it

mrc0mmand:
Some distributions have started phasing out SHA1, which breaks the systemd-measure test case in its current form. Let's make sure we can use SHA1 for signing beforehand to mitigate this.

Spotted on RHEL 9, where SHA1 signatures are disallowed by [0]:
```
openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -out "/tmp/pcrsign-private.pem"
...
openssl rsa -pubout -in "/tmp/pcrsign-private.pem" -out "/tmp/pcrsign-public.pem"
writing RSA key
/usr/lib/systemd/systemd-measure sign --current --bank=sha1 --private-key="/tmp/pcrsign-private.pem" --public-key="/tmp/pcrsign-public.pem"
Failed to initialize signature context.
```

[0] https://gitlab.com/redhat/centos-stream/rpms/openssl/-/blob/c9s/0049-Selectively-disallow-SHA1-signatures.patch

(cherry picked from commit d19e5540f20c78caa949ff33050b4a530cae1982)

Related: #2141979

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -1572,7 +1572,7 @@ static EFI_STATUS efivar_get_timeout(const char16_t *var, uint32_t *ret_value) {
 
 static void config_load_defaults(Config *config, EFI_FILE *root_dir) {
         _cleanup_free_ char *content = NULL;
-        UINTN value;
+        UINTN value = 0;  /* avoid false maybe-uninitialized warning */
         EFI_STATUS err;
 
         assert(root_dir);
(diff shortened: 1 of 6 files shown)


mrc0mmand:
Waiting for https://github.com/systemd/systemd/issues/25641.

dtardon (approved):
LGTM

PR #124, opened 2022-12-08
(#2138081) More `chase_symlinks*()` shenanigans

mrc0mmand:
This reverts & backports the new set of patches for the recent `chase_symlinks*()` without `/proc/` issue.

For RHEL we either want to backport https://github.com/systemd/systemd/pull/25661 as well once it's merged, or drop the warning completely/lower it to debug, hence opening this as a draft for now.

/cc @msekletar


msekletar:
@jamacku we need to wait with this one until https://github.com/systemd/systemd/pull/25661 is sorted out. On top of that we will need to change rpm macros and add `--no-warn` to systemctl invocations.

msekletar (approved):
@mrc0mmand thanks for following up on this, much appreciated!

LGTM

PR #132, opened 2023-01-12
(#2161260)  resolvectl dns does not display localhost as Global DNS server

jacekmigacz:
resolvectl: don't filter loopback DNS server from global DNS server list
    
    "resolvectl status" shows per-link DNS servers separately from global
    ones. When querying the global list, it will contain both per-link and
    global servers however. Thus, to not show duplicate info we filter all
    entries that actually have a non-zero ifindex set (under the assumption
    that that's a per-link server).
    
    This doesn't work if people configured 127.0.0.1 as global server
    though, as we'll add ifindex 1 to it since
    6e32414a66ff8dbcef233981a7066684d903ee9f unconditionally even for global
    servers.
    
    Let's address that by excluding entries with ifindex 1 from suppression.
    This is safe as resolved ignores loopback ifaces, hence never will have
    per-link servers on ifindex 1.
    
    Note that this splits up the "with_ifindex" parameter into a second
    parameter "only_global", since they semantically do two different
    things. One controls whether we shall expect/parse an ifindex dbus
    field. The other controls whether we shall filter all ifindex values set
    != 0. These are effectively always used in conjunction hence making them
    the same actually worked. However this is utterly confusing I think,
    which as I guess is resulting in the confusion around #25796 (which
    removes the whole check)

### DIFF
--- a/src/resolve/resolvectl.c
+++ b/src/resolve/resolvectl.c
@@ -1203,9 +1203,10 @@ static int reset_server_features(int argc, char **argv, void *userdata) {
 static int read_dns_server_one(sd_bus_message *m, bool with_ifindex, bool extended, char **ret) {
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
         _cleanup_free_ char *pretty = NULL;
-        int ifindex, family, r, k;
         union in_addr_union a;
         const char *name = NULL;
+        int32_t ifindex = 0;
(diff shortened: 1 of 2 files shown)


jamacku (changes requested):
@jacekmigacz, Bug [#2148870](https://bugzilla.redhat.com/show_bug.cgi?id=2148870) is for Fedora, but you need RHEL9 bug.

PR #135, opened 2023-02-02
(#2138081) Assorted post-v252 backports

mrc0mmand:


test/units/testsuite-75.sh
```diff
@@ -212,6 +212,11 @@ resolvectl log-level debug
 
 # Start monitoring queries
 systemd-run -u resmontest.service -p Type=notify resolvectl monitor
+# Wait for the monitoring service to become active
+for _ in {0..9}; do
```
dtardon: There really should be something like `systemctl wait` there...

test/units/testsuite-75.sh
```diff
@@ -41,7 +41,7 @@ monitor_check_rr() (
     # displayed. We turn off pipefail for this, since we don't care about the
     # lhs of this pipe expression, we only care about the rhs' result to be
     # clean
-    journalctl -u resmontest.service --since "$since" -f --full | grep -m1 "$match"
+    timeout -v 30s journalctl -u resmontest.service --since "$since" -f --full | grep -m1 "$match"
```
dtardon, line 44: Isn't this an unnecessary complication? You already have `--since`, hence `-f` shouldn't even be needed, should it?

dtardon (approved):
LGTM

test/units/testsuite-75.sh
mrc0mmand: The `--since` is a workaround for an issue (?) with journal timestamps, see https://github.com/redhat-plumbers/systemd-rhel9/pull/135/commits/f664e6699d5e06aadf0ab6ec1e44333b58d9d9db for more details (we need `-f`, since we need to wait for the message to appear, and `--since=` to make sure we don't miss any relevant entries).

test/units/testsuite-75.sh
mrc0mmand: Yeah... thankfully this hunk has been dropped a couple of commits later.

PR #138, opened 2023-02-13
(#1952378) core: when isolating to a unit, also keep units running that are trig…

dtardon:
…gered by units we keep running

Inspired by: #26364

(this might even "fix" #26364, but without debug logs it's hard to make such claims)

Fixes: #23055
(cherry picked from commit 32d6707dd1692d41e12f5469dfdcbc10f14d6619)

Resolves: [#1952378](https://bugzilla.redhat.com/show_bug.cgi?id=1952378)

### DIFF
--- a/src/core/transaction.c
+++ b/src/core/transaction.c
@@ -1092,6 +1092,20 @@ fail:
         return r;
 }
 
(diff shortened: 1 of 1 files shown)


msekletar (approved):
LGTM

PR #139, opened 2023-02-16
(#2170500) udev/net_id: introduce naming scheme for RHEL-9.2

jamacku:
RHEL-only

Resolves: [#2170500](https://bugzilla.redhat.com/show_bug.cgi?id=2170500)


dtardon (approved):
LGTM

mrc0mmand:
Ignore the CentOS CI fails - the QEMU workaround I put in place doesn't work anymore, I'll need to test a couple of things to make it work again. As the changes in this PR are quite straightforward, it doesn't make sense to block it until I fix the CIs.

src/shared/netif-naming-scheme.h
```diff
@@ -53,6 +53,7 @@ typedef enum NamingSchemeFlags {
         NAMING_V252 = NAMING_V251 | NAMING_DEVICETREE_ALIASES,
         NAMING_RHEL_9_0 = NAMING_V250 | NAMING_BRIDGE_MULTIFUNCTION_SLOT,
         NAMING_RHEL_9_1 = NAMING_RHEL_9_0,
+        NAMING_RHEL_9_2 = NAMING_RHEL_9_0,
```
dtardon, line 56: Eh, this should have been set to `NAMING_V252`. But I don't think it's worth changing it now, as the "devicetree aliases" thingy doesn't look very important.

jamacku: I believe that we have discussed it with @msekletar and concluded that we don't want to change the naming scheme mid-major release. But I might be mistaken.

dtardon: `rhel-9.0` shall remain _the default_ scheme, sure. But subsequent schemes should cover changes. That's the whole point of their existence.

PR #151, opened 2023-03-14
(#2178222) Revert workaround for disabled idmapped mounts

dtardon:
Resolves: [#2178222](https://bugzilla.redhat.com/show_bug.cgi?id=2178222)


msekletar (approved):
LGTM

mrc0mmand:
Could you do a force-push to restart the CIs? The results are so old that they're no longer accessible.

PR #153, opened 2023-03-17
(#2179165) unit: In cgroupv1, gracefully terminate delegated scopes again

jamacku:
Instantly killing delegated scopes is just not viable for our needs in OCP 4.13 in cgroupv1 mode.  We will accept the possibility of timeouts instead.

Co-authored-by: Colin Walters <walters@verbum.org>

rhel-only

Resolves: #2179165


jamacku:
Let's merge this one so we can start testing.

PR #161, opened 2023-04-25
ci(Mergify): configuration update

jamacku:
This change has been made by @jamacku from Mergify config editor.


lnykryn (dismissed):

dtardon:
What if we introduced a new label that would be set only if a PR changes code paths checked by CodeQL and then made the Mergify check for CodeQL results conditional on that label?

jamacku:
I wonder why the Commit Linter GA couldn't detect the commit id from the body of the first comment. I will have a look.

@dtardon Thanks for the feedback. I have added a condition for changed files. The same problem is with build workflows, which run only when the source code is changed. Please have a look if it looks fine. Thanks

.mergify.yml
```diff
@@ -9,11 +9,15 @@ pull_request_rules:
       - label!=ci-waived
       - or:
         # Build test
-        - -check-success=build (gcc, 11, bfd, gcrypt)
-        - -check-success=build (gcc, 12, gold, openssl)
-        - -check-success=build (clang, 13, mold, gcrypt) 
-        - -check-success=build (clang, 14, lld, openssl)
-        - -check-success=build (clang, 15, bfd, auto)
+        # test build check only when specific files are changed ...
```
dtardon, line 12: Maybe "test build check only when the tests have been run" or something like that?

.mergify.yml
```diff
@@ -9,11 +9,15 @@ pull_request_rules:
...
+        - and:
+          - files~=^(\S+\/meson\.build|\.github\/workflows\/\S+|meson_options\.txt|src\/\S+|test\/fuzz\/\S+)
```
dtardon, line 14: Maybe add a comment to the build/codeql config files saying: "If changed, also change ... in `.mergify.yml`"? That way, if there's any need to update the paths in the future, these regexes won't get out of sync.

dtardon (commented):
LGTM, with minor suggestions.

.mergify.yml
jamacku: Unfortunately, this is currently not possible. [Available conditions](https://docs.mergify.com/conditions/#attributes):

Screenshot from 2023-04-27 18-24-53

.mergify.yml
jamacku: Addressed in https://github.com/redhat-plumbers/systemd-rhel9/pull/161/commits/55c11ce8fd7ad9ce46d212504ed2262db2fee454

.mergify.yml
jamacku: Sorry, I misunderstood. I have updated the comments.

PR #162, opened 2023-05-09
(#2193456) manager: don't taint the host if cgroups v1 is used

jamacku:
In upstream of systemd, cgroups v1 are not considered as supported. This is not true for RHEL, don't taint the host when cgroups v1 are enabled.

rhel-only

Resolves: #2193456

### DIFF
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -1590,12 +1590,6 @@ node /org/freedesktop/systemd1 {
         </varlistentry>
 
         <varlistentry>
-          <term><literal>cgroupsv1</literal></term>
-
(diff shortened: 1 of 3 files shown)


msekletar (approved):
LGTM

mrc0mmand:
Same issue as in https://github.com/redhat-plumbers/systemd-rhel9/pull/163#issuecomment-1633867753.

jamacku:
> I think you should also remove the mention of `cgroupsv1` taint flag from `man/org.freedesktop.systemd1.xml`, so the man page is consistent with the code.

Good point, thank you.

dtardon (approved):
LGTM

PR #163, opened 2023-05-09
(#2196479) manager: don't taint the host if cgroups v1 is used

jamacku:
In upstream of systemd, cgroups v1 are not considered as supported. This is not true for RHEL, don't taint the host when cgroups v1 are enabled.

rhel-only

Resolves: #2196479

### DIFF
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -1590,12 +1590,6 @@ node /org/freedesktop/systemd1 {
         </varlistentry>
 
         <varlistentry>
-          <term><literal>cgroupsv1</literal></term>
-
-          <listitem><para>The system is using the old cgroup hierarchy.</para></listitem>
-        </varlistentry>
-
-        <varlistentry>
           <term><literal>local-hwclock</literal></term>
 
           <listitem><para>The local hardware clock (RTC) is configured to be in local time rather than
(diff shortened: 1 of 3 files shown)


msekletar (dismissed):
LGTM

msekletar:
@mrc0mmand it would be great if we could introduce internal test for this. To make sure we don't forget about this for RHEL-10.

mrc0mmand:
> @mrc0mmand it would be great if we could introduce internal test for this. To make sure we don't forget about this for RHEL-10.

I was kind of (naively) hoping we'd get rid of cgroupsv1 stuff in RHEL 10, especially since it's going to be removed completely in upstream next year: https://github.com/systemd/systemd/blob/076deab75d1b8371dd6709341d6bea69b42bf7d2/NEWS#L14-L19

mrc0mmand:
```
 440/1402 systemd:test / test-manager                                                             FAIL              0.39s   killed by signal 6 SIGABRT
>>> PATH=/root/systemd/build:/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin SYSTEMD_LANGUAGE_FALLBACK_MAP=/root/systemd/src/locale/language-fallback-map MALLOC_PERTURB_=100 /root/systemd/build/test-manager
stderr:
/* test_manager_taint_string */
taint string w/o split-usr: ''
Found cgroup on /sys/fs/cgroup/systemd, legacy hierarchy
Assertion 'strstr(a, "cgroupsv1")' failed at src/test/test-manager.c:18, function test_manager_taint_string(). Aborting.
```

PR #166, opened 2023-05-17
(RHEL-6090) Add new Type=notify-reload service type for a service reload protocol based on SIGHUP 

dtardon:
Resolves: [#2207781](https://bugzilla.redhat.com/show_bug.cgi?id=2207781)


msekletar (changes requested):
L(mostly)GTM but there is this one more bugfix we must add otherwise we might hit some fairly difficult to debug issues.

https://github.com/systemd/systemd/commit/845824acddf2e7e08c94afe7cfee6e50a682c947

jamacku (dismissed):
@dtardon, the original bug was migrated to JIRA, so, unfortunately, we have to reference instead:

https://issues.redhat.com/browse/RHEL-6090

jamacku (commented):
LGTM

msekletar (approved):
LGTM

PR #173, opened 2023-05-30
(#2211065) Always use inactive_exit_timestamp if it is set

dtardon:
Resolves: [#2211065](https://bugzilla.redhat.com/show_bug.cgi?id=2211065)


msekletar (approved):
LGTM

PR #175, opened 2023-06-08
(#2213521) (#2213521) Make unit dependency always bidirectional

dtardon:
Resolves: #2213521


mrc0mmand:
> `CentOS CI (CentOS Stream 9)` fails in `TEST-75-RESOLVED`. I don't think that's related.

Yup, definitely not related. I think it's related to the resolved issue discussed internally, but we see this in the upstream CIs as well.

msekletar (approved):
LGTM

PR #176, opened 2023-06-21
(#2212612) rules: online CPU automatically on IBM s390x platforms

legoater:
RHEL-only

Fix CPU hotplug regression on s390x introduced by commit 94c7e260b499 ("rules: do not online CPU automatically on IBM platforms")

Resolves: #2212612

### DIFF
--- a/rules.d/40-redhat.rules
+++ b/rules.d/40-redhat.rules
@@ -3,7 +3,8 @@
 # CPU hotadd request
 SUBSYSTEM!="cpu", GOTO="cpu_online_end"
 ACTION!="add", GOTO="cpu_online_end"
-CONST{arch}=="s390*|ppc64*", GOTO="cpu_online_end"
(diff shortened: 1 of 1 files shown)


sharkcz:
AFAIK this will try to set online the stand-by CPUs that can't be onlined :-( Ideally the rule should check also the "configurable" attribute, see https://bugzilla.redhat.com/show_bug.cgi?id=2006257 for details.

legoater:
> AFAIK this will try to set online the stand-by CPUs that can't be onlined :-( Ideally the rule should check also the "configurable" attribute, see https://bugzilla.redhat.com/show_bug.cgi?id=2006257 for details.

Yes. After discussing with the IBM Z folks, hotplug CPUs should be automatically onlined only if they are configured. The rule lacks a check on the sysfs  `configure` attribute.

legoater:
Just fixed a typo. CPU hotplug should work as requested now.

msekletar (approved):
LGTM

PR #178, opened 2023-06-28
(#2210145) pam: add call to pam_umask

jamacku:
Setting umask for user sessions via UMASK setting in /etc/login.defs is a well-known feature. Let's make sure that user manager also runs with this umask value.

Follow-up for 5e37d1930b41b24c077ce37c6db0e36c745106c7.

(cherry picked from commit 159f1b78576ce91c3932f4867f07361a530875d3)

Resolves: #2210145


msekletar (approved):
LGTM

PR #181, opened 2023-07-10
(#2170883) Allow `RHEL-only` labels to mark downstream-only commits

jamacku:


jamacku:
Note: Updated configuration needs to be available on the `main` branch for this to work. It's the standard behavior of `on: workflow_run` trigger.

dtardon (approved):
LGTM

PR #182, opened 2023-07-11
(#2218886) udev-builtin-net_id: align VF representor names with VF names

dtardon:
Certain cards support to set their eswitch to switchdev mode. In this mode for each created VF there is also created so called VF representor. This representor is helper network interface used for configuration of mentioned eswitch and belongs to an appropriate PF.

VF representors are identified by the specific value of phys_port_name attribute and the value has format "pfMvfN" where M is PF function number and N is VF number inside this PF.

As the VF representor interfaces belong to PF PCI device the naming scheme used for them is the same like for other PCI devices. In this case name of PF interface is used and phys_port_name suffix is appended.

E.g.
 PF=enp65s0f0np0 # phys_port_name for PF interface is 'p0'
 VF=enp65s0f0np0v0 # v0 is appended for VF0 in case of NAMING_SR_IOV_V
REP=enp65s0f0np0pf0vf0 # phys_port_name for VF0 representor is 'pf0vf0'

First as the phys_port_name for representors is long (6+ chars) then the generated name does not fit into IFNAMSIZ so this name is used only as alternate interface name and for the primary one is used generic one like eth<N>. Second 'f0' and 'pf0' in REP name is redundant.

This patch fixes this issue by introducing another naming scheme for VF representors and appending 'rN' suffix to PF interface name for them. N is VF number so the name used for representor interface is similar to VF interface and differs only by the suffix.

For the example above we get:
 PF=enp65s0f0np0
 VF=enp65s0f0np0v0
REP=enp65s0f0np0r0

This eases for userspace to determine which representor interface represents particular VF.

Signed-off-by: Ivan Vecera <ivecera@redhat.com>
(cherry picked from commit 88d2bda8120dcc375a90e28b64de06b9646ab3b6)

Resolves: [#2218886](https://bugzilla.redhat.com/show_bug.cgi?id=2218886)


msekletar (approved):
LGTM

PR #184, opened 2023-07-13
(#2170883) ci: drop systemd-stable from advanced-commit-linter config

jamacku:
It's sufficient enough to check only the `systemd/systemd` repo.

Related to https://github.com/redhat-plumbers-in-action/advanced-commit-linter/issues/62

rhel-only

Related: #2170883


dtardon (approved):
LGTM

PR #186, opened 2023-08-01
Revert "core/service: when resetting PID also reset known flag"

jamacku:
This reverts commit ff32060f2ed37b68dc26256b05e2e69013b0ecfe.

This change is incorrect as we don't want to mark the PID as invalid but only mark it as dead.

The change in question also breaks user level socket activation for `podman.service` as the termination of the main `podman system service` process is not properly handled, causing any application accessing the socket to hang.

This is because the user-level `podman.service` unit also hosts two non-main processes: `rootlessport` and `rootlessport-child` which causes the `cgroup_good` check to still succeed.

The original submitter of this commit is recommended to find another more correct way to fix the cgroupsv1 issue on CentOS 8.

(cherry picked from commit 996b00ede87d6a870332e63974a7d4def3c2f1b0)

Resolves: #2225667
Reverts: #2210237

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -86,7 +86,7 @@ if [[ "$COMPILER" == clang ]]; then
                "$RELEASE" "$RELEASE" "$COMPILER_VERSION" >/etc/apt/sources.list.d/llvm-toolchain.list
     fi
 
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Ad the failing CIs: https://github.com/systemd/systemd/commit/c5afbac31bb33e7b1f4d59b253425af991a630a4 should fix that.

msekletar (approved):
LGTM

PR #189, opened 2023-08-08
(#2170883) doc: improve CONTRIBUTING document

jamacku:
based on feedback from: https://github.com/redhat-plumbers/systemd-rhel8/pull/407

rhel-only

Related: #2170883


dtardon (approved):
LGTM

PR #191, opened 2023-08-08
(RHEL-5988) Assign alternative names only on add uevent

dtardon:

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -1108,6 +1108,25 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
         if (action == SD_DEVICE_MOVE)
                 device_remove_old_on_move(m, dev);
 
+        /* When udevd failed to process the device, SYSTEMD_ALIAS or any other properties may contain invalid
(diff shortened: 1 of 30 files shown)


msekletar:
@dtardon CI failures looks legitimate.

lnykryn:
The shellcheck failure is a false positive. https://github.com/koalaman/shellcheck/wiki/SC2317#exceptions

jamacku (dismissed):
@dtardon, the original bug was migrated to JIRA, so, unfortunately, we have to reference instead:

https://issues.redhat.com/browse/RHEL-5988

msekletar (approved):
LGTM

mrc0mmand:
The test introduced in this PR actually fails (as pointed out by the CI), and it causes the CI to fail in other PRs as well (https://github.com/redhat-plumbers/systemd-rhel9/pull/215#issuecomment-1847886782).

mrc0mmand:
Ah, found the missing commits, will prep a PR shortly.

mrc0mmand:
-> https://github.com/redhat-plumbers/systemd-rhel9/pull/216

PR #192, opened 2023-08-10
(#2170883) docs: link downstream CONTRIBUTING in README

jamacku:
This should increase the visibility of the downstream CONTRIBUTING.

Also, fix some wording and update links.

rhel-only

Related: #2170883


dtardon (approved):
LGTM

PR #193, opened 2023-08-10
(#2230364) bpf: fix restrict_fs on s390x

jamacku:
Linux kernel's bpf-next contains BPF LSM support for s390x. systemd's test-bpf-lsm currently fails with this kernel.

This is an endianness issue: in the restrict_fs bpf program, magic_number has type unsigned long (64 bits on s390x), but magic_map keys are uint32_t (32 bits). Accessing magic_map using 64-bit keys may work by accident on little-endian systems, but fails hard on big-endian ones.

Fix by casting magic_number to uint32_t.

(cherry picked from commit 907046282c27ee2ced5e22abb80ed8df2e157baf)

Resolves: #2230364


dtardon (approved):
LGTM

PR #195, opened 2023-08-14
(#2231845) udev/net_id: use naming scheme for RHEL-9.3

jamacku:
rhel-only

Resolves: #2231845

https://github.com/redhat-plumbers/systemd-rhel9/blob/eff259c8c7677bbbe64890873bd64349871dd6c5/man/systemd.net-naming-scheme.xml#L486-L495

https://github.com/redhat-plumbers/systemd-rhel9/blob/eff259c8c7677bbbe64890873bd64349871dd6c5/src/shared/netif-naming-scheme.h#L58


lnykryn (approved):

dtardon (approved):
Oops... LGTM

PR #198, opened 2023-09-22
(RHEL-7026) netif-naming-scheme: let's also include rhel8 schemes

lnykryn:
With this patch user in rhel9 can also pick a scheme from rhel8. This could be useful on in-place upgrades.

rhel-only
Resolves: RHEL-7026


jamacku (changes requested):
Could you please also add documentation [man/systemd.net-naming-scheme.xml](https://github.com/redhat-plumbers/systemd-rhel9/blob/main/man/systemd.net-naming-scheme.xml)

Similarly to: https://github.com/redhat-plumbers/systemd-rhel9/blob/55d337de1940076855c1687ffd588498d068724e/man/systemd.net-naming-scheme.xml#L462-L495

lnykryn:
@jamacku done

jamacku (approved):
LGTM

PR #201, opened 2023-09-29
(#2239576) rules: online CPU automatically on IBM s390x platforms when configured

bmasonrh:
RHEL-only

Fix CPU hotplug regression on s390x introduced by commit 94c7e260b499 ("rules: do not online CPU automatically on IBM platforms"). After discussion with IBM, CPUs should be auto-enabled when in the configured state after a hotplug. However, if the CPU is deconfigured, it should not.

This is because on zVM and KVM hotplugged CPUs are configured and on LPAR/DPM they are deconfigured.

Resolves: #2239576
Signed-off-by: Cédric Le Goater <clg@redhat.com>

(cherry picked from commit 65d993c2efe52d683396dc3181cc79f29698bf39)

### DIFF
--- a/rules.d/40-redhat.rules
+++ b/rules.d/40-redhat.rules
@@ -3,7 +3,8 @@
 # CPU hotadd request
 SUBSYSTEM!="cpu", GOTO="cpu_online_end"
 ACTION!="add", GOTO="cpu_online_end"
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM

bmasonrh:
Not sure why, but it seems like this MR was still in draft status.  I've changed it to Open so you should be able to merge it now.

PR #211, opened 2023-11-22
(RHEL-13199) Backport ukify from systemd-255-rc2

keszybz:
This is just the first piece. I'm starting with this to let the CI run. We would also need to update the spec file to use the new option, and figure out how to deal with the `pefile` dependency. But I think this should be handled in the `dist-git` repo.

All the tests pass on Fedora 39.


keszybz:
I didn't include the patch that moves `ukify` to `/usr/bin/`… Do we want that here?

jamacku:
@msekletar I think that this two commits are an exception. It would be impossible to bacport individual commits a reference them correctly. In the same time these commits aren't, a rhel-only because that would indicate that we want to keep them I case of rebase. 

I think it's fine to leave them as they are.

jamacku:
Maybe we should have some label for this case like we have the rhel-only label. On the other hand we don't have this situation often.

msekletar:
@jamacku Fine by me then.

PR #212, opened 2023-11-23
(RHEL-16354) UKI support in kernel-install

keszybz:

### DIFF
--- a/src/boot/bootctl-uki.c
+++ b/src/boot/bootctl-uki.c
@@ -30,10 +30,15 @@ static int pe_sections(FILE *uki, struct PeSectionHeader **ret, size_t *ret_n) {
         assert(ret_n);
 
         items = fread(&dos, 1, sizeof(dos), uki);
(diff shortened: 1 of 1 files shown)


jamacku (dismissed):
@keszybz Please update commit message of commit - 506ff04af8be4fb189c9524fe1330853599b7df1- kernel-install: handle uki installs automatically.

Please replace `Resolves: #25822` with `Resolves: https://github.com/systemd/systemd/issues/25822`.

`Resolves: #` is a keyword used in RHEL to reference Bugzilla bugs. And since we don't allow multiple trackers in one PR, our automation is failing.

msekletar (approved):
LGTM

msekletar:
Btw, bootctl pieces weren't strictly needed (I think) as we don't really use systemd-boot in RHEL but I guess it doesn't hurt to have them in as bootctl is useful to some extent even without systemd-boot.

PR #215, opened 2023-12-08
(RHEL-13199) meson: fix installation of ukify

keszybz:
ln_s was added in upstream later on. It's not present in this branch. Fixup for b98da2d9e7.

Related: RHEL-13199

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3868,9 +3868,9 @@ ukify = custom_target(
 if want_ukify
         public_programs += ukify
 
-        meson.add_install_script(sh, '-c',
-                                 ln_s.format(bindir / 'ukify',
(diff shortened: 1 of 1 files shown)


mrc0mmand:
```
00:08:16.347 Message: Skipping ukify.1 because ENABLE_UKIFY is false
```

Ah, that's why CIs didn't catch it. Lemme fix that.

mrc0mmand:
```
 353/1406 systemd:ukify / test-ukify                                                              OK                0.71s
```

Yup, now it seems to run the tests as well. The TEST-17-UDEV fail in the sanitizers run is unrelated, though concerning, I'll look into that later.

msekletar (approved):
LGTM

PR #216, opened 2023-12-11
(RHEL-5988) `udevadm trigger --settle` follow-ups

mrc0mmand:
Addresses https://github.com/redhat-plumbers/systemd-rhel9/pull/191#issuecomment-1849740706 where `udevadm trigger --settle` freezes when used with an about-to-be-renamed device, see https://github.com/systemd/systemd/issues/25115.


mrc0mmand:
The last commit is from https://github.com/redhat-plumbers/systemd-rhel9/pull/215 to make the build pass, once that PR is merged I'll rebase this one.

msekletar (approved):
LGTM

PR #217, opened 2023-12-14
(RHEL-19436) man: mention RHEL documentation in systemctl's man page

mrc0mmand:
The original link pointed to a non-existing page, so let's update it.


jamacku (approved):

dtardon (approved):
LGTM

PR #218, opened 2023-12-19
(RHEL-1317) udev: allow/denylist for reading sysfs attributes when composing a NIC name

lnykryn:
Users can currently pick specific versions of NIC naming, but that does not guarantee that NIC names won't change after the kernel adds a new sysfs attribute.

This patch allows for an allow/deny list of sysfs attributes that could be used when composing the name.

These lists can be supplied as an hwdb entry in the form of /etc/udev/hwdb.d/50-net-naming-allowlist.hwdb
net:naming:drvirtio_net
  ID_NET_NAME_ALLOW=0
  ID_NET_NAME_ALLOW_ACPI_INDEX=1
  ID_NET_NAME_ALLOW_ADDR_ASSIGN_TYPE=1
  ID_NET_NAME_ALLOW_ADDRESS=1
  ID_NET_NAME_ALLOW_ARI_ENABLED=1
  ID_NET_NAME_ALLOW_DEV_PORT=1
  ID_NET_NAME_ALLOW_FUNCTION_ID=1
  ID_NET_NAME_ALLOW_IFLINK=1
  ID_NET_NAME_ALLOW_INDEX=1
  ID_NET_NAME_ALLOW_LABEL=1
  ID_NET_NAME_ALLOW_PHYS_PORT_NAME=1
  ID_NET_NAME_ALLOW_TYPE=1

(cherry picked from commit 3b2e7dc5a285edbbb1bf6aed2d88b889d801613f)

Resolves: RHEL-1317

### DIFF
--- a/man/systemd.net-naming-scheme.xml
+++ b/man/systemd.net-naming-scheme.xml
@@ -601,7 +601,7 @@
       <varlistentry>
         <term><varname>ID_NET_NAME_ALLOW=<replaceable>BOOL</replaceable></varname></term>
 
-        <listitem><para>This evironment value sets a fallback policy for reading a sysfs attribute.
+        <listitem><para>This udev property sets a fallback policy for reading a sysfs attribute.
         If set to <constant>0</constant> udev will not read any sysfs attribute by default, unless it is
         explicitly allowlisted, see below. If set to <constant>1</constant> udev can use any sysfs attribute
         unless it is explicitly forbidden. The default value is <constant>1</constant>.</para>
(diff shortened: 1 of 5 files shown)


msekletar (approved):
LGTM

Btw, while reviewing this I realized it would be kind of nice to enforce use of `*_filtered()` function variants in `src/udev/udev-builtin-net_id.c`. 

@mrc0mmand do you think it would be possible to write CodeQL query that would enforce this policy?

PR #220, opened 2024-01-02
(RHEL-11591) Avoid SIGBUS during journal file rotation

dtardon:
Resolves: RHEL-11591


mrc0mmand:
Is the issue reference here correct? Since https://issues.redhat.com/browse/RHEL-11591 is for RHEL 9.2, but this PR is opened against current main.

mrc0mmand:
> Is the issue reference here correct? Since https://issues.redhat.com/browse/RHEL-11591 is for RHEL 9.2, but this PR is opened against current main.

I'm an idiot, nevermind, looked at the wrong field :)

mrc0mmand (approved):
LGTM.

jamacku:
> > Is the issue reference here correct? Since https://issues.redhat.com/browse/RHEL-11591 is for RHEL 9.2, but this PR is opened against current main.
> 
> I'm an idiot, nevermind, looked at the wrong field :)

No :) This should be improved. The text is confusing, but it's partially due to JIRA because it doesn't have a field that would tell you what major version of RHEL is affected. :(

PR #225, opened 2024-01-17
(RHEL-19483) core: mount namespaces: Remove auxiliary bind mounts directory after …

lnykryn:
…unit termination

Unit that requires its own mount namespace creates a temporary directory to implement dynamic bind mounts (org.freedesktop.systemd1.Manager.BindMountUnit). However, this directory is never removed and they will accumulate for each unique unit (e.g. templated units of systemd-coredump@).

Attach the auxiliary runtime directory existence to lifetime of other "runtime" only per-unit directories.

(cherry picked from commit b9f976fb45635e09cd709dbedd0afb03d4b73c05)

Resolves: RHEL-19483

### DIFF
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -5478,6 +5478,23 @@ int exec_context_destroy_credentials(const ExecContext *c, const char *runtime_p
         return 0;
 }
 
+int exec_context_destroy_mount_ns_dir(Unit *u) {
+        _cleanup_free_ char *p = NULL;
+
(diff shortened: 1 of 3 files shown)


dtardon (approved):
LGTM

jamacku:
@mrc0mmand, could you please check CentOS CI? Thank you.

mrc0mmand:
The test fail is a known bug we probably won't fix in RHEL 9 (https://github.com/systemd/systemd/issues/24452).

PR #227, opened 2024-01-22
(RHEL-22167) man: remove mentions of `systemd.network(5)`

jamacku:
This is an attempt to remove all the references to `systemd.network(5)`, as we don't ship that file in RHEL9.

rhel-only

Resolves: RHEL-22167

### DIFF
--- a/man/systemd-resolved.service.xml
+++ b/man/systemd-resolved.service.xml
@@ -81,8 +81,7 @@
     is used), the per-link dynamic settings received over DHCP, information provided via
     <citerefentry><refentrytitle>resolvectl</refentrytitle><manvolnum>1</manvolnum></citerefentry>, and any
     DNS server information made available by other system services. See
-    <citerefentry><refentrytitle>resolved.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry> and
-    <citerefentry><refentrytitle>systemd.network</refentrytitle><manvolnum>5</manvolnum></citerefentry> for
+    <citerefentry><refentrytitle>resolved.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry> for
     details about systemd's own configuration files for DNS servers. To improve compatibility,
     <filename>/etc/resolv.conf</filename> is read in order to discover configured system DNS servers, but
     only if it is not a symlink to <filename>/run/systemd/resolve/stub-resolv.conf</filename>,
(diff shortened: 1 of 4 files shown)


man/systemd-resolved.service.xml
```diff
@@ -409,7 +407,6 @@ search foobar.com barbar.com
       <citerefentry><refentrytitle>resolvectl</refentrytitle><manvolnum>1</manvolnum></citerefentry>,
       <citerefentry project='man-pages'><refentrytitle>resolv.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
       <citerefentry project='man-pages'><refentrytitle>hosts</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
-      <citerefentry><refentrytitle>systemd.network</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
       <citerefentry><refentrytitle>systemd-networkd.service</refentrytitle><manvolnum>8</manvolnum></citerefentry>
```
dtardon, line 410: If we're removing mentions of `systemd.network` then mentions of `systemd-networkd.service` should be dropped as well.

man/systemd.link.xml
```diff
@@ -1238,9 +1238,6 @@ MACAddress=cb:a9:87:65:43:21</programlisting>
       <citerefentry>
         <refentrytitle>systemd.netdev</refentrytitle><manvolnum>5</manvolnum>
```
dtardon, line 1239: Also `systemd.netdev` and `systemd-network-generator.service`.

dtardon (commented):
IMHO this is a pointless exercise (I'm pretty sure there are other man pages in RHEL that refer to man pages or binaries _not_ present in RHEL), but whatever...

man/systemd-resolved.service.xml
jamacku: Yes, there are a lot more issues as I've mentioned here: [RHEL-22167#comment-23844272](https://issues.redhat.com/browse/RHEL-22167?focusedId=23844272&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23844272).

As you mentioned, it is a pointless exercise, so I would only remove mentions of `systemd.network` and close the issue. @dtardon What do you think?

dtardon: I replied on the issue. Let's wait a bit...

PR #229, opened 2024-01-23
(RHEL-22443) udev: add new builtin net_driver

lnykryn:
Currently the ID_NET_DRIVER is set in net_setup_link builtin. But this is called pretty late in the udev processing chain.

Right now in some custom rules it was workarounded by calling ethtool binary directly, which is ugly.

So let's split this code to a separate builtin.

(cherry picked from commit 2b5b25f123ceb89b3ff45b2380db1c8a88b046d9)

Resolves: RHEL-22443


lnykryn:
For the reviewer, in rhel9 we don't have ERRNO_IS_NEG_NOT_SUPPORTED. I've used ERRNO_IS_NOT_SUPPORTED instead. The first one should only eat negative errnos, while the old one eats both negative and positives errnos. But I don't think this is an issue.  ethtool_get_driver always returns either negative value or zero.

PR #237, opened 2024-02-26
(RHEL-26643) resolved: limit the number of signature validations in a transaction

jacekmigacz:


jamacku (changes requested):
@jacekmigacz We usually don't reference merge commits. Please reference the following upstream commits instead:

- [resolved: limit the number of signature validations in a transaction](https://github.com/systemd/systemd/commit/67d0ce8843d612a2245d0966197d4f528b911b66)
- [resolved: reduce the maximum nsec3 iterations to 100](https://github.com/systemd/systemd/commit/eba291124bc11f03732d1fc468db3bfac069f9cb)

msekletar (approved):
LGTM

PR #240, opened 2024-02-29
udev: raise RLIMIT_NOFILE as high as we can

dtardon:
We might need a lot of fds on large systems, hence raise RLIMIT_NOFILE to what the service manager allows us, which is quite a lot these days.

udev already sets FORK_RLIMIT_NOFILE_SAFE when forking of chilren, thus ensuring that forked off processes get their RLIMIT_NOFILE soft limit reset to 1K for compat with crappy old select().

Replaces: #29298
Fixes: #28583
(cherry picked from commit 1617424ce76d797d081dd6cb1082b954c4d2bf38)

Resolves: RHEL-27382

### DIFF
--- a/src/udev/udevd.c
+++ b/src/udev/udevd.c
@@ -55,6 +55,7 @@
 #include "pretty-print.h"
 #include "proc-cmdline.h"
 #include "process-util.h"
+#include "rlimit-util.h"
 #include "selinux-util.h"
 #include "signal-util.h"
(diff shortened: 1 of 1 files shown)


jamacku:
@brozs, @bmasonrh - This one is probably for you :)

mrc0mmand:
Slapped on two commits that should, hopefully, fix the CIs.

PR #238, opened 2024-02-26
(RHEL-27512) backport new mkosi

lnykryn:
I am not sure if we want this, but I really like using mkosi when working upstream. But mkosi from f39 simply does not work with systemd-rhel9.

So here is a PoC with a rebased configuration.

* Works on my machine
* Quite a lot of tests are failing

So right now I am able to make it work with

```
mkosi genkey
mkosi -fT  -d centos -r 9 qemu

```

PR #241, opened 2024-02-29
(RHEL-16182) units: change assert to condition to skip running in initrd/os

mrc0mmand:
These units are also present in the initrd, so instead of an assert, just use a condition so they are skipped where they need to be skipped.

Fixes https://github.com/systemd/systemd/issues/26358

(cherry picked from commit 7ef09e2099a4f97ad40748d6b7c735b45aa4c990)

Related: RHEL-16182

---

Noticed this while testing #238. This doesn't affect RHEL 9 directly (yet), since RHEL 9's dracut doesn't pull in these files, but once that change gets backported we'd start seeing fails in journal. Since we're going to do another RHEL 9 build anyway, let's throw this one in as well.

### DIFF
--- a/units/systemd-pcrfs-root.service.in
+++ b/units/systemd-pcrfs-root.service.in
@@ -14,7 +14,7 @@ DefaultDependencies=no
 Conflicts=shutdown.target
 After=systemd-pcrmachine.service
 Before=shutdown.target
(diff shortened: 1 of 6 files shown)


dtardon (approved):
LGTM

PR #242, opened 2024-02-29
(RHEL-26133) efi: alignment of the PE file has to be at least 512 bytes

msekletar:
https://learn.microsoft.com/en-us/windows/win32/debug/pe-format?redirectedfrom=MSDN#optional-header-windows-specific-fields-image-only

Resolves: RHEL-26133

RHEL-only

[msekleta: this is RHEL-only because upstream no longer uses objcopy to create PE files]


dtardon (approved):
LGTM

PR #254, opened 2024-04-09
(RHEL-32259) execute: Pass AT_FDCWD instead of -1

lnykryn:
Let's enforce that callers pass AT_FDCWD as read_dfd to load_credential() to avoid an assert() in read_full_file_full() if read_dfd is -1.

(cherry picked from commit 661e4251a5b157d1aee1df98fbd2f0c95285ebba)

Resolves: RHEL-32259


dtardon (approved):
LGTM

PR #257, opened 2024-04-10
Fix `journalctl --follow` with non-matching filter

dtardon:
Also fixes a bunch of other `journalctl` issues with combinations of options.

Resolves: RHEL-30567


msekletar:
Hi @dtardon, it seems that followup detection is correct. Can you please double-check and backport missing followups if appropriate. Also, some commits are missing JIRA reference. Thank you!

msekletar:
Build failures in CI are relevant and should be fixed before we can move forward with this.

PR #261, opened 2024-04-24
(RHEL-29430) coredump: generate stacktraces also for processes running in containers w/o coredump forwarding

msekletar:
Note that entering container namespace has to be explicitly enabled by setting SYSTEMD_COREDUMP_ALLOW_NAMESPACE_CHANGE environment variable.

RHEL-only

Resolves: RHEL-29430

### DIFF
--- a/src/basic/socket-util.c
+++ b/src/basic/socket-util.c
@@ -41,6 +41,11 @@
 #  define IDN_FLAGS 0
 #endif
 
+/* From the kernel's include/net/scm.h */
+#ifndef SCM_MAX_FD
+#  define SCM_MAX_FD 253
+#endif
+
 static const char* const socket_address_type_table[] = {
         [SOCK_STREAM] =    "Stream",
         [SOCK_DGRAM] =     "Datagram",
@@ -949,6 +954,53 @@ int getpeergroups(int fd, gid_t **ret) {
         *ret = TAKE_PTR(d);
 
         return (int) n;
+}
+
+ssize_t send_many_fds_iov_sa(
+                int transport_fd,
+                int *fds_array, size_t n_fds_array,
+                const struct iovec *iov, size_t iovlen,
+                const struct sockaddr *sa, socklen_t len,
+                int flags) {
+
+        _cleanup_free_ struct cmsghdr *cmsg = NULL;
+        struct msghdr mh = {
+                .msg_name = (struct sockaddr*) sa,
+                .msg_namelen = len,
+                .msg_iov = (struct iovec *)iov,
+                .msg_iovlen = iovlen,
+        };
+        ssize_t k;
+
+        assert(transport_fd >= 0);
+        assert(fds_array || n_fds_array == 0);
+
+        /* The kernel will reject sending more than SCM_MAX_FD FDs at once */
+        if (n_fds_array > SCM_MAX_FD)
+                return -E2BIG;
+
+        /* We need either an FD array or data to send. If there's nothing, return an error. */
+        if (n_fds_array == 0 && !iov)
+                return -EINVAL;
+
+        if (n_fds_array > 0) {
+                mh.msg_controllen = CMSG_SPACE(sizeof(int) * n_fds_array);
+                mh.msg_control = cmsg = malloc(mh.msg_controllen);
+                if (!cmsg)
+                        return -ENOMEM;
+
+                *cmsg = (struct cmsghdr) {
+                        .cmsg_len = CMSG_LEN(sizeof(int) * n_fds_array),
+                        .cmsg_level = SOL_SOCKET,
+                        .cmsg_type = SCM_RIGHTS,
+                };
+                memcpy(CMSG_DATA(cmsg), fds_array, sizeof(int) * n_fds_array);
+        }
+        k = sendmsg(transport_fd, &mh, MSG_NOSIGNAL | flags);
+        if (k < 0)
+                return (ssize_t) -errno;
+
+        return k;
 }
 
 ssize_t send_one_fd_iov_sa(
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -24,6 +24,7 @@
 #include "coredump-vacuum.h"
 #include "dirent-util.h"
 #include "elf-util.h"
+#include "env-util.h"
 #include "escape.h"
 #include "fd-util.h"
 #include "fileio.h"
@@ -36,6 +37,7 @@
 #include "main-func.h"
 #include "memory-util.h"
 #include "mkdir-label.h"
+#include "namespace-util.h"
 #include "parse-util.h"
 #include "process-util.h"
 #include "signal-util.h"
@@ -130,6 +132,8 @@ typedef struct Context {
         const char *meta[_META_MAX];
         size_t meta_size[_META_MAX];
         pid_t pid;
+        uid_t uid;
+        gid_t gid;
         bool is_pid1;
         bool is_journald;
 } Context;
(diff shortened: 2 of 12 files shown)


src/coredump/coredump.c
```diff
@@ -1192,16 +1202,16 @@ static int process_socket(int fd) {
...
         iovw_free_contents(&iovw, true);
         return r;
 }
 
-static int send_iovec(const struct iovec_wrapper *iovw, int input_fd) {
-        _cleanup_close_ int fd = -1;
-        int r;
+static int send_iovec(const struct iovec_wrapper *iovw, int input_fd, int mntns_fd) {
+        _cleanup_close_ int fd = -EBADF;
+         int r;
```
mrc0mmand, line 1214: Spurious indentation.

src/shared/elf-util.c
```diff
@@ -811,6 +835,23 @@ int parse_elf_object(int fd, const char *executable, bool fork_disable_dump, cha
...
+                                            /* netns_fd = */ -EBADF,
+                                            /* userns_fd = */ -EBADF,
+                                            /* root_fd = */ -EBADF);
+                        if (r < 0)
+                                log_notice_errno(r, "Failed to enter mount namespace of crashing process, ignoring: %m");
+                }
+
+                if (uid != UID_NOBODY && gid != GID_NOBODY) {
+                        r = core_change_uid_gid(uid, gid);
+                        if (r < 0) {
```
mrc0mmand, line 850: No {} around single-line blocks.

src/coredump/coredump.c
```diff
@@ -1462,6 +1478,33 @@ static int process_kernel(int argc, char* argv[]) {
                 log_open();
         }
 
+        r = in_same_namespace(getpid_cached(), context.pid, NAMESPACE_PID);
+        if (r < 0)
+                log_debug_errno(r, "Failed to check pidns of crashing process, ignoring: %m");
+
+        if (r == 0 && getenv_bool("SYSTEMD_COREDUMP_ALLOW_NAMESPACE_CHANGE") > 0) {
```
mrc0mmand, line 1485: I'm trying to ~~break~~ write some tests for this and I'm slightly confused _where_ this env variable is supposed to be set? It can't be set in the `systemd-coredump@.service` as that's too late (that service is spawned once the `kernel.core_pattern` sd-coredump instance is done processing the coredump), it won't propagate to the `core_pattern` binary environment from the manager, and setting it in the crashed process' environment doesn't work either, since the `core_pattern` binary is executed outside of that environment. So in all cases this `getenv()` will return `-ENXIO`.

msekletar: Well, for I testing I pointed to core_pattern to my own script which sets env variable and just executes systemd-coredump (this is roughly what actually what customer is going to do as well).

msekletar: Here is what I used,

```
$ sysctl kernel.core_pattern
kernel.core_pattern = |/usr/local/bin/coredumper %P %u %g %s %t %c %h

$ cat /usr/local/bin/coredumper
#!/bin/bash

export SYSTEMD_COREDUMP_ALLOW_NAMESPACE_CHANGE=1
exec /usr/lib/systemd/systemd-coredump "$@"
```

src/coredump/coredump.c
```diff
@@ -1146,23 +1136,39 @@ static int process_socket(int fd) {
...
+                                safe_close(input_fd);
+
+                                input_fd = fds[0];
+                                mntns_fd = fds[1];
+
+                                /* We have all FDs we need let's take a shortcut here. */
+                                break;
+                        } else {
+                                found = cmsg_find(&mh, SOL_SOCKET, SCM_RIGHTS, CMSG_LEN(sizeof(int)));
+                                if (found)
```
dtardon, line 1164: You've lost an `assert()` here.

src/basic/socket-util.c
```diff
@@ -951,6 +956,53 @@ int getpeergroups(int fd, gid_t **ret) {
...
+        /* The kernel will reject sending more than SCM_MAX_FD FDs at once */
+        if (n_fds_array > SCM_MAX_FD)
+                return -E2BIG;
+
+        /* We need either an FD array or data to send. If there's nothing, return an error. */
+        if (n_fds_array == 0 && !iov)
+                return -EINVAL;
+
+        if (n_fds_array > 0) {
+                mh.msg_controllen = CMSG_SPACE(sizeof(int) * n_fds_array);
```
dtardon, line 987: Nitpicking, but maybe save `sizeof(int) * n_fds_array` to a variable?

src/coredump/coredump.c
```diff
@@ -1462,6 +1478,33 @@ static int process_kernel(int argc, char* argv[]) {
...
+                        return log_error_errno(r, "Failed to open mntns of crashing process: %m");
+
+                mntns_fd = fd;
+        } else {
+                /* Crashing process is not running in the container or changing namespace is disabled, but we
+                   still need to send mount namespace fd along side coredump fd so let's just open our own
+                   mount namespace. Entering it will be NOP but that is OK. */
+                int fd;
+
+                r = namespace_open(getpid_cached(),  NULL, &fd, NULL, NULL, NULL);
```
dtardon, line 1499: Same here.

src/coredump/coredump.c
```diff
@@ -1462,6 +1478,33 @@ static int process_kernel(int argc, char* argv[]) {
...
+
+                r = namespace_open(context.pid,  NULL, &fd, NULL, NULL, NULL);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to open mntns of crashing process: %m");
+
+                mntns_fd = fd;
+        } else {
+                /* Crashing process is not running in the container or changing namespace is disabled, but we
+                   still need to send mount namespace fd along side coredump fd so let's just open our own
+                   mount namespace. Entering it will be NOP but that is OK. */
```
dtardon, line 1496: Why? The ELF parsing code seems to handle `mntns_fd` being unset.

src/coredump/coredump.c
msekletar: Actually I left it out deliberately. Previously, we asserted that this is the first fd attached to 0 length message because input_fd < 0 and we stored it to the input_fd variable and broke out from loop. Now we potentially do multiple iterations so it seemed to me that we are violating "spirit" of that assertion anyway.

src/coredump/coredump.c
```diff
@@ -1462,6 +1478,33 @@ static int process_kernel(int argc, char* argv[]) {
...
         }
 
+        r = in_same_namespace(getpid_cached(), context.pid, NAMESPACE_PID);
+        if (r < 0)
+                log_debug_errno(r, "Failed to check pidns of crashing process, ignoring: %m");
+
+        if (r == 0 && getenv_bool("SYSTEMD_COREDUMP_ALLOW_NAMESPACE_CHANGE") > 0) {
+                int fd;
+
+                r = namespace_open(context.pid,  NULL, &fd, NULL, NULL, NULL);
```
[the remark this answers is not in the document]
msekletar: No reason, maybe readability but that is questionable. I will fix it. Thanks!

src/coredump/coredump.c
msekletar: I wanted to keep special casing limited to this only. So the code which is sending fds would work the same way in both cases.

msekletar: But if that is a blocker for you I can change it.

dtardon: Not a blocker. It just confused me, as I'd expected this would only be needed for the special case.

dtardon (approved):
LGTM

mrc0mmand:
I slapped a couple of tests commit on top, to at least make sure the coredump machinery is not borked. It would be great if this coverage got into the z-stream clones as well. It looks like we got TEST-74-AUX-UTILS in 9.2.0 onward, so backporting it there should be theoretically easy. For older releases we'd have to trust the patch I guess.

mrc0mmand:
(Also a CI note - the CI will currently fail until #264 and #260 are merged, but TEST-74-AUX-UTILS should, hopefully, pass).

dtardon (approved):
Still LGTM

PR #262, opened 2024-04-24
(RHEL-34061) coredump: generate stacktraces also for processes running in containers w/o coredump forwarding

msekletar:
Note that entering container namespace has to be explicitly enabled by setting SYSTEMD_COREDUMP_ALLOW_NAMESPACE_CHANGE environment variable.

RHEL-only

Resolves: RHEL-34061


jamacku (approved):
LGTM

PR #267, opened 2024-05-14
(RHEL-35703) manager: restrict Dump*() to privileged callers or ratelimit

lnykryn:

### DIFF
--- a/test/units/testsuite-65.sh
+++ b/test/units/testsuite-65.sh
@@ -51,6 +51,21 @@ systemd-analyze dot --require systemd-journald.service systemd-logind.service >/
 systemd-analyze dot "systemd-*.service" >/dev/null
 (! systemd-analyze dot systemd-journald.service systemd-logind.service "*" bbb ccc)
 # dump
+# this should be rate limited to 10 calls in 10 minutes for unprivileged callers
+for _ in {1..10}; do
+    runas testuser systemd-analyze dump systemd-journald.service >/dev/null
+done
+(! runas testuser systemd-analyze dump >/dev/null)
+# still limited after a reload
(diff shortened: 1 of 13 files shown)


lnykryn:
The failure from FuzzLite seems to be unrelated, but it should be fixed with https://github.com/redhat-plumbers/systemd-rhel9/pull/269

dtardon (approved):
LGTM

jamacku:
@mrc0mmand Could you please have a look at CentOS CI? Thank you :pray:

test/units/testsuite-65.sh
```diff
@@ -51,6 +51,21 @@ systemd-analyze dot --require systemd-journald.service systemd-logind.service >/
 systemd-analyze dot "systemd-*.service" >/dev/null
 (! systemd-analyze dot systemd-journald.service systemd-logind.service "*" bbb ccc)
 # dump
+# this should be rate limited to 10 calls in 10 minutes for unprivileged callers
```
mrc0mmand, line 54: We don't have the `test/units/util.sh` script in RHEL9, so you'll need to copy the definition of `runas()` here:

https://github.com/systemd/systemd/blob/e3828d7103a99a15a1e947ba3063294ead590631/test/units/util.sh#L133C1-L137C2

msekletar (approved):

test/units/testsuite-65.sh
jamacku: addressed in https://github.com/redhat-plumbers/systemd-rhel9/pull/267/commits/27e836bd3eacb25d0a714cf59b2272316670ae26

PR #275, opened 2024-05-23
(RHEL-6589) libsystemd: link with '-z nodelete'

msekletar:
We want to avoid reinitialization of our global variables with static storage duration in case we get dlopened multiple times by the same application. This will avoid potential resource leaks that could have happened otherwise (e.g. leaking journal socket fd).

Resolves: RHEL-6589


dtardon (approved):
LGTM

PR #282, opened 2024-06-18
(RHEL-30372) fuzz: pass -Dc_args=/-Dcpp_args= to fuzzer targets

mrc0mmand:
Prompted by #29972, because right now it's practically impossible to pass
-fno-sanitize=function to the fuzzer targets without some extensive
sed'ing.

This splits both c_args and cpp_args to separate arguments for
tools/meson-build.sh, because the other way would be to use `eval`, so
the space-separated but quoted strings passed to these options are not
split where they shouldn't, and I'd rather avoid using `eval` if
possible.

Also, this switches the positional arguments we pass to `meson setup`,
as they were in incorrect order (docs say it should be buildir followed
by sourcedir); meson is apparently clever enough to figure this out and
switch the arguments around if necessary, so it didn't complain.

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel9/commit/17ee59c9c922553a8cb4d54cb8ae415706c4feff)

Related: RHEL-30372

---

Apart from the reason in the original commit, I need this for some internal shenanigans as well.

### DIFF
--- a/test/fuzz/meson.build
+++ b/test/fuzz/meson.build
@@ -38,10 +38,11 @@ sanitize_address_undefined = custom_target(
                    project_source_root,
                    '@OUTPUT@',
                    'fuzzers',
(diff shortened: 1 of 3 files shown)


jamacku (approved):
LGTM

PR #283, opened 2024-07-01
(RHEL-44630) netif-naming-scheme: disable NAMING_BRIDGE_MULTIFUNCTION_SLOT

lnykryn:
This effectively reverts 66425daf2c68793adf24a48a26d58add8662e83f.

The commit assumes that if the network interface has multifunctions, then the function fields of the pci devices under the same PCI bridge device are unique.
But it seems not, at least on some setups. See issue #28929. Let's revert the change, and always refuse to set slot base naming if a PCI bridge is detected.

Fixes #28929.

(cherry picked from commit af7417ac7b07bc01232982bf46e9d72e69e7f820)

Resolves: RHEL-44630

### DIFF
--- a/man/systemd.net-naming-scheme.xml
+++ b/man/systemd.net-naming-scheme.xml
@@ -499,6 +499,20 @@
 
            <listitem><para>Same as naming scheme <constant>rhel-9.3</constant>.</para></listitem>
         </varlistentry>
+
+        <varlistentry>
+            <term><constant>rhel-9.5</constant></term>
+
+          <listitem><para>Never generate a slot name when a PCI bridge is detected.</para>
+
+          <para>Since version <constant>"rhel-9.0"</constant>, we have generated slot-based names
+          for PCI multifunction devices, because we assumed that it is enough to use function numbers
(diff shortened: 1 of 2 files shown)


msekletar (changes requested):
I don't like the wording of the manpage paragraph. It is difficult to understand in my opinion so I've tried to rewrite it for better clarity

```
Never generate a slot name when a PCI bridge is detected.

Since version "rhel-9.0", we have generated slot-based names for PCI multifunction devices, because we assumed that it is enough to use function numbers to distinguish between devices. However, name conflict can occur if these devices are not children of the same PCI bridge, e.g. there are multiple PCI bridges in the same slot.
```

msekletar (approved):
LGTM

champtar:
You are missing `{ "rhel-9.5", NAMING_RHEL_9_5 }` in the .c

lnykryn:
> You are missing `{ "rhel-9.5", NAMING_RHEL_9_5 }` in the .c

Good catch thanks! https://github.com/redhat-plumbers/systemd-rhel9/pull/284

PR #284, opened 2024-07-03
(RHEL-44630) netif-naming-scheme: make actually possible to use rhel-9.5 scheme

lnykryn:
In 753e1b1c9b255d528eb8b2a2af072a83eb85d784 we forgot to update the netif-naming-scheme.c file to allow people to actually use the new rhel-9.5 scheme.

RHEL-only

Resolves: RHEL-44630


msekletar (approved):
LGTM

PR #285, opened 2024-07-16
(RHEL-33436) generator: "uninline" generator_open_unit_file and generator_add_symlink

msekletar:
Inlining of these functions changed ABI of libsystemd-shared which causes issue on update when generators packaged in systemd-udev subpackage fail to execute because of ABI change. systemd and the library are already updated while systemd-udev subpackage is not and hence old generators can't be started when systemd is reexecuting due to internal library incompatibility.

rhel-only: bugfix

Resolves: RHEL-33436

### DIFF
--- a/src/shared/generator.c
+++ b/src/shared/generator.c
@@ -75,6 +75,10 @@ int generator_open_unit_file_full(
         return 0;
 }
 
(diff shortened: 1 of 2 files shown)


msekletar:
FWIW, I tested it locally and the patch seems to fix the issue.

dtardon (approved):
LGTM

PR #287, opened 2024-07-17
(RHEL-49495) generator: "uninline" generator_open_unit_file and generator_add_symlink

jamacku:
Inlining of these functions changed ABI of libsystemd-shared which causes issue on update when generators packaged in systemd-udev subpackage fail to execute because of ABI change. systemd and the library are already updated while systemd-udev subpackage is not and hence old generators can't be started when systemd is reexecuting due to internal library incompatibility.

rhel-only: bugfix

Resolves: RHEL-49495

### DIFF
--- a/src/shared/generator.c
+++ b/src/shared/generator.c
@@ -109,6 +109,10 @@ int generator_add_symlink_full(
         return 0;
 }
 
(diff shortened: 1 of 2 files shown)


msekletar (approved):
LGTM

PR #288, opened 2024-07-22
(RHEL-45020) timedatectl: setting set_local_rtc to 1 will throw Warning as well, u…

dtardon:
…se log_warning() (#33489)

Previously only running `timedatectl` it was showing warning regarding the dangers of setting RTC to local TZ.
Now similar warning is also flashed when `set-local-rtc 1`.

(cherry picked from commit db708fc9ba551325e000cc12bd3490a5de0715ba)

Resolves: RHEL-45020


jamacku (approved):
LGTM

PR #293, opened 2024-08-06
(RHEL-5950) Update HWDB

dtardon:
Resolves: RHEL-5950


jamacku:
Apparently, the automation can't handle so many commits, I'll try to have a look, but for now, I'll do a manual verification.

@msekletar, Could you please have a look at the changes? Thank you

jamacku (approved):
This is huge, but it's touching only HWDB stuff, and it seems OK. LGTM

PR #295, opened 2024-08-22
(RHEL-50103) Revert "udev-builtin-net_id: use firmware_node/sun for ID_NET_NAME_SLOT"

lnykryn:
This reverts commit 7af151ca282d506a8e409a68b656f6d1cd2f13fb.

It seems that virtio devices always have "0" in
the firmware_node/sun. And because of that, udev will always name the device ens0, which leads to collisions. So let's disable it for now.

rhel-only: policy

Reverts: RHEL-50103


msekletar (approved):
LGTM

dtardon:
By a coincidence, a fix for this issue was proposed--and approved--on the same day: https://github.com/systemd/systemd/pull/34099 ...

PR #298, opened 2024-09-10
repart: avoid use of uninitialized TPM2B_PUBLIC data

dtardon:
The 'TPM2B public' struct is only initialized if the public key is non-NULL, however, it is unconditionally passed to tpm2_calculate_sealing_policy, resulting in use of uninitialized data. If the uninitialized data is lucky enough to be all zeroes, this results eventually results in an error message from tpm2_calculate_name about an unsupported nameAlg field value.

Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>
(cherry picked from commit a3ad5c3140b941d3703c63c902e58f4e2d295829)

Resolves: RHEL-56793


jamacku:

```
```

PR #299, opened 2024-09-10
(RHEL-55301) Don't GC unit if it is in another queue

dtardon:
Resolves: RHEL-55301

### DIFF
--- a/src/core/cgroup.c
+++ b/src/core/cgroup.c
@@ -3007,7 +3007,9 @@ static int on_cgroup_empty_event(sd_event_source *s, void *userdata) {
 
         unit_add_to_gc_queue(u);
 
-        if (UNIT_VTABLE(u)->notify_cgroup_empty)
+        if (IN_SET(unit_active_state(u), UNIT_INACTIVE, UNIT_FAILED))
+                unit_prune_cgroup(u);
+        else if (UNIT_VTABLE(u)->notify_cgroup_empty)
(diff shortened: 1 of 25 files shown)


msekletar (changes requested):
I think that github action warning is a real issue which should be fixed.

jamacku:
> I think that github action warning is a real issue which should be fixed.

We need https://github.com/systemd/systemd/commit/cb153b4fe9046590eb8efba66b67a0db899cb96d or we have to use the original name `test/units/assert.sh`.

https://github.com/systemd/systemd/commits/cb153b4fe9046590eb8efba66b67a0db899cb96d/test/units/assert.sh?browsing_rename_history=true&new_path=test/units/util.sh&original_branch=main

dtardon:
> > I think that github action warning is a real issue which should be fixed.
> 
> We need [systemd/systemd@cb153b4](https://github.com/systemd/systemd/commit/cb153b4fe9046590eb8efba66b67a0db899cb96d) or we have to use the original name `test/units/assert.sh`.
> 
> https://github.com/systemd/systemd/commits/cb153b4fe9046590eb8efba66b67a0db899cb96d/test/units/assert.sh?browsing_rename_history=true&new_path=test/units/util.sh&original_branch=main

I backported the rename.

jamacku:
CI failure is unrelated

```
systemd-stub@0x67d7b000,0x67d93000
```

msekletar (approved):
LGTM

PR #301, opened 2024-10-03
(RHEL-36531) man: using WantedBy=default.target is not a good idea

lnykryn:
We had several users, that wrote their unit files with WantedBy=default.target because it should be started "every time". But for example in Fedora/CentOS/RHEL, this often breaks for example selinux relabels (where we just want to do a relabel and reboot).

(cherry picked from commit 67b6404b80cf8078f3d9ec6d4c2f34ac25b15077)

Resolves: RHEL-36531


jamacku (approved):
LGTM

PR #303, opened 2024-11-19
(RHEL-24340) logind: tighten for which classes of sessions we do stop-on-idle

msekletar:
We only want to do this for fully set up, interactive sessions, i.e. user and user-early, but not for any others, hence restrict the rules a bit.

Follow-up for: 508b4786e8592e82eb4832549f74aaa54335d14c

(cherry picked from commit ad23439eae718ac3634f260be0d29e01445983a8)

Related: RHEL-24340


jamacku:
Thank you @mrc0mmand

PR #306, opened 2024-12-20
(RHEL-5956) install: don't translate unit instances to paths when reenabling them

jamacku:
For unit instances install_info_discover() returns path to the template, which then generates confusing errors when passed to do_unit_file_enable():

~# build/systemctl --root=/tmp/systemctl-test.N9ysbz reenable templ1@two.service Unit name: templ1@two.service; p: /etc/systemd/system/templ1@.service Removed "/tmp/systemctl-test.N9ysbz/etc/systemd/system/services.target.wants/templ1@two.service". Failed to reenable templ1@.service, destination unit services.target is a non-template unit.

This can also be seen with a different reproducer using getty@.service and a simple bind mount to / - there's no error this time, but it tries to create a symlink for the default instance (from DefaultInstance=tty1), which is also incorrect:

~# SYSTEMD_LOG_LEVEL=debug systemctl --root /mnt/bindroot/ reenable getty@test.service Symlink /mnt/bindroot/etc/systemd/system/getty.target.wants/getty@tty1.service → /usr/lib/systemd/system/getty@.service already exists

Follow-up to: 29a7c59abbe
Resolves upstream issue #24740

(cherry picked from commit fe6e0cfa19dd1de4ac599ae207182fd556adcfa7)

Resolves: RHEL-5956

### DIFF
--- a/src/shared/install.c
+++ b/src/shared/install.c
@@ -2875,7 +2875,7 @@ static int normalize_linked_files(
                         return log_debug_errno(SYNTHETIC_ERRNO(EISDIR),
                                                "Unexpected path to a directory \"%s\", refusing.", *a);
 
-                if (!is_path(*a)) {
+                if (!is_path(*a) && !unit_name_is_valid(*a, UNIT_NAME_INSTANCE)) {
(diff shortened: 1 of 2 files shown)


dtardon (approved):
LGTM

jamacku:

```
```

PR #307, opened 2024-12-20
(RHEL-33815) udev-node: optimize device node symlink creation

jamacku:

### DIFF
--- a/src/udev/udev-node.c
+++ b/src/udev/udev-node.c
@@ -104,7 +104,8 @@ static int node_symlink(sd_device *dev, const char *devnode, const char *slink) 
         return 0;
 }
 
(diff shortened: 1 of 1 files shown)


msekletar:
Honestly, udev symlink handling is just a mess. It is best to not change it in already released RHEL as we have no idea what might break. However, I some cases it is necessary, I think we can merge this but at the beginning of next development cycle.

msekletar:
I've cleared the milestone, now we need to look at mentions and potential followup and really think through all potential implications of these changes and potentially merge early at the beginning of RHEL-9.7 devdoctest cycle.

msekletar:
Very are way past the point I feel comfortable with such risky backports so let's keep the status quo for now. Also both cases related to this PR are closed.

PR #310, opened 2025-01-06
(RHEL-3906) systemctl: print warning about missing /proc only as debug message

msekletar:
rhel-only

Resolves: RHEL-3906


jamacku (commented):
Plain `rhel-only` is no longer supported. Please use `rhel-only` with one of the keywords from - https://github.com/redhat-plumbers/.github/blob/main/docs/SYSTEMD_CONTRIBUTING.md#upstream-reference

jamacku:

```
```

jamacku (approved):
LGTM

PR #311, opened 2025-01-14
(RHEL-70884) ci: use ubuntu 22:04 for deploy of man pages

jamacku:
rhel-only: ci

Related: RHEL-70884


jamacku:
CI failures are unrelated to the change.

PR #314, opened 2025-01-31
(RHEL-15051) man: be even clearer that tmpfiles user/group/mode are applied on exi…

jamacku:
…sting inodes

I think it was clear already, but let's be even clearer.

Fixes: #29774
(cherry picked from commit 3cb938bd12b3603984b982e9b73e4cabd4a608e3)

Resolves: RHEL-15051


dtardon (approved):
LGTM

PR #315, opened 2025-03-14
coredump: check mnt ns equality to decide if we shall change our mnt ns.

Romain-Geissler-1A:
Indeed there is no reason that the pid namespace is not shared with thehost in a container. However by definition the mount namespace isn't shared. Typically in OpenShift, when we "oc debug" a node, we get a new container (so a dedicated mount namespace) but this container shares the host pid namespace. Core dumps happening in such a container have a wrong backtrace reported by systemd-coredumper as it doesn't even try to change its mount namspace while it should.

Thus it seems more correct to compare mount namspaces rather than pid namespaces.

@msekletar This is related to our call from last week and the Red Hat case I will send you by mail.

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -1478,9 +1478,9 @@ static int process_kernel(int argc, char* argv[]) {
                 log_open();
         }
 
-        r = in_same_namespace(getpid_cached(), context.pid, NAMESPACE_PID);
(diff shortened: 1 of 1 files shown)


msekletar:
Hi @Romain-Geissler-1A,

thank you for the patch. Problematic piece of code was in upstream already at the time when I was adding EnterNamespace= so we need to take this discussion upstream first. It would be great if you could submit this change against github.com/systemd/systemd or let me know if you want me to do it. Thanks!

PR #319, opened 2025-04-07
(RHEL-79977) ci: enable mkosi workflow for z-stream branches

jamacku:
rhel-only: ci

Related: RHEL-79977


dtardon (approved):
Superficially LGTM

PR #321, opened 2025-04-09
(RHEL-68299) logind: tighten for which classes of sessions we do stop-on-idle

bmasonrh:
We only want to do this for fully set up, interactive sessions, i.e. user and user-early, but not for any others, hence restrict the rules a bit.

Follow-up for: 508b4786e8592e82eb4832549f74aaa54335d14c

(cherry picked from commit ad23439eae718ac3634f260be0d29e01445983a8)

Resolves: RHEL-68299

PR #322, opened 2025-04-09
(RHEL-86237) unit: always return 1 in log_kill

bmasonrh:
This ensures that cg_kill_items returns the correct value to let the manager know that a process was killed.

(cherry picked from commit 500cd2e83b8246fbf20d99db898039cfba746223)

Resolves: RHEL-86237


jamacku (commented):
The CI failure is nonblocking and is the same problem as in #321.

msekletar (approved):
LGTM

PR #324, opened 2025-04-14
man: mention that rhel-9.0 net naming scheme is the same as v251

lnykryn:
Due to the rebase in 9.2 it is a bit confusing that we are documenting the newer upstream scheme and it might look 9.0 is based in v255. So let's explicitly mention that the scheme is the same as in v251.

Resolves: RHEL-86891

rhel-only: doc


jamacku (approved):
LGTM

PR #325, opened 2025-04-14
(RHEL-86714) backlight: Drop support for actual_brightness

jamacku:
Some AMD systems have support for features like custom brightness curve or adaptive backlight management.  These features allow the display driver to adjust the brightness based upon other factors than just the user brightness request.

The user's brightness request is indicated in the 'brightness' file but the effective result of the logic in the display driver is stored in the 'actual_brightness' file.

This leads to problems when shutting the system down because the value of 'actual_brightness' may be lower than 'brightness' and the wrong value gets stored for the next boot.

For example if the brightness a user requested was 150, the actual_brightness might be 130. So the next boot the brightness will be "set" to 130, but the actual brightness might be 115. If the user reboots again it will be set to 115 for the next boot but the actual brightness might be 100. That is this gets worse and worse each reboot cycle until the system eventually boots up at minimum brightness.

Furthermore the kernel documentation indicates that the brightness and actual_brightness files are not guaranteed to be the same values.

Due to this; drop the use of 'actual_brightness' when saving/restoring brightness and instead rely only upon 'brightness'.

Signed-off-by: Mario Limonciello <mario.limonciello@amd.com>

(cherry picked from commit 9a224c307b36610e3675390eb2620e74d0f4efb0)

Resolves: RHEL-86714

### DIFF
--- a/src/backlight/backlight.c
+++ b/src/backlight/backlight.c
@@ -386,46 +386,13 @@ static bool shall_clamp(sd_device *d) {
 }
 
 static int read_brightness(sd_device *device, unsigned max_brightness, unsigned *ret_brightness) {
-        const char *subsystem, *value;
+        const char *value;
         unsigned brightness;
         int r;
 
         assert(device);
         assert(ret_brightness);
 
-        r = sd_device_get_subsystem(device, &subsystem);
-        if (r < 0)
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

PR #327, opened 2025-04-17
(RHEL-85467) user-sessions: do not remove /etc/nologin

lnykryn:
pam_nologin looks for /etc/nologin and /run/nologin. user-sessions creates (and removes) /run/nologin, but also removes /etc/nologin. (This behaviour is unchanged since the introduction of the binary in e92787416c691c3f34f47349e5eae3fa68eae856.)

By not removing pam_nologin we fully drop compatibility with PAM < 1.1. This has the advantage that now /etc/nologin can be used by administrator to disable user logins, e.g. for extended maintanance. We already specified PAM >= 1.1.2 as dependency, so this was already covered.

The makes the code match the man page.

Fixes #26965.

(cherry picked from commit a78413baae0e999384b535d327203ebf417b1e24)

Resolves: RHEL-85467

### DIFF
--- a/src/shared/fileio-label.c
+++ b/src/shared/fileio-label.c
@@ -23,9 +23,15 @@ int write_string_file_atomic_label_ts(const char *fn, const char *line, struct t
 int create_shutdown_run_nologin_or_warn(void) {
         int r;
 
-        /* This is used twice: once in systemd-user-sessions.service, in order to block logins when we actually go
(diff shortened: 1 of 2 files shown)


dtardon (approved):
LGTM

msekletar (approved):
LGTM

jamacku:
CI failures are unrelated and known issues.

PR #331, opened 2025-06-12
(RHEL-25516) udev/net_id: introduce naming scheme for RHEL-9.7

jamacku:
rhel-only: policy

Resolves: RHEL-25516


dtardon (approved):
LGTM

PR #333, opened 2025-06-27
flush_ports: flush POSIX message queues properly

dtardon:
On Linux, read() on a message queue descriptor returns the message queue statistics, not the actual message queue data.  We need to use mq_receive() to drain the queues instead.

Fixes a problem where a POSIX message queue socket unit with messages in the queue at shutdown time could result in a hang on reboot/shutdown.

(cherry picked from commit ffb6adb76367d5ab7d43937ccaac5947717b5b78)

Resolves: RHEL-90860

### DIFF
--- a/src/basic/socket-util.c
+++ b/src/basic/socket-util.c
@@ -3,6 +3,7 @@
 #include <arpa/inet.h>
 #include <errno.h>
 #include <limits.h>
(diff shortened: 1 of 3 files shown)


jamacku:
mkosi failure is a known issue

PR #334, opened 2025-07-11
(RHEL-29761) fix: do not check/verify slice units if recursive errors are to be ignored

bmasonrh:
Before this fix, when recursive-errors was set to 'no' during a systemd-analyze verification, the parent slice was checked regardless. The 'no' setting means that, only the specified unit should be looked at and verified and errors in the slices should be ignored. This commit fixes that issue.

Example:

Say we have a sample.service file:

[Unit]
Description=Sample Service

[Service]
ExecStart=/bin/echo "a"
Slice=support.slice

Before Change:

systemd-analyze verify --recursive-errors=no maanya/sample.service Assertion 'u' failed at src/core/unit.c:153, function unit_has_name(). Aborting. Aborted (core dumped)

After Change:
systemd-analyze verify --recursive-errors=no maanya/sample.service {No errors}

(cherry picked from commit f660c7fa56b247c278fdb2ebcfea37912f249524)

Related: RHEL-29761

### DIFF
--- a/.github/workflows/gather-metadata.yml
+++ b/.github/workflows/gather-metadata.yml
@@ -22,7 +22,7 @@ jobs:
         uses: redhat-plumbers-in-action/gather-pull-request-metadata@v1
 
       - name: Upload artifact with gathered metadata
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: pr-metadata
           path: ${{ steps.Metadata.outputs.metadata-file }}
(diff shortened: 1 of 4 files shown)


jamacku:
@bmasonrh to fix GitHub Actions and Packit CI you can also backport following commits:

- https://github.com/redhat-plumbers/systemd-rhel9/commit/e95a2756c94f9bcf6283cd899d6cbb0e048a181d
- https://github.com/redhat-plumbers/systemd-rhel9/commit/06a9ab5d2fc8ddd0223d91b9a5ce2685d49a5b18

PR #338, opened 2025-07-29
(RHEL-100353) Turn ordering cycle log message into a recognizable one + catalog entry

dtardon:
Resolves: RHEL-100353


jamacku:
@dtardon We may also want this commit: https://github.com/systemd/systemd/commit/1c8d653d2b8a82ed3944f03c2dea25ad1e2cc967

dtardon:
> @dtardon We may also want this commit: [systemd/systemd@1c8d653](https://github.com/systemd/systemd/commit/1c8d653d2b8a82ed3944f03c2dea25ad1e2cc967)

Right, that makes sense. I'll also create a new PR for RHEL10 with it, as the "main" one has already been merged.

PR #339, opened 2025-08-05
(RHEL-103753)  Drop inotify event from buffer when no event source is triggered

dtardon:
Resolves: RHEL-103753


mrc0mmand:
The CI fail is expected and should be resolved once #340 is merged.

PR #343, opened 2025-08-13
(RHEL-103488) core: ensure init.scope is realized after drop-ins have been loaded

dtardon:
If we add a drop-in for init.scope (e.g.: to set some memory limit), it will be loaded long after the cgroup has already been realized. Do it again when creating the special unit.

(cherry picked from commit 020b2e41ea776cff73392da8084a0725b590d245)

Resolves: RHEL-103488


jamacku (approved):
LGTM, ci failures are unrelated to this change.

PR #353, opened 2025-08-28
(RHEL-111611) meson: /etc/systemd/network is also used by udevd

jamacku:
(cherry picked from commit 6256c65aad2a719ac9054961561bb26e497208ce)

Resolves: RHEL-111611


dtardon (approved):
LGTM

PR #355, opened 2025-08-28
(RHEL-108587) compress: replace compress_blob() with compress_blob_explicit()

jamacku:
And make compress_xyz() return 0 on success, as we know which compression algorithm is used when calling compress_blob().

Follow-up for 2360352ef02548723ac0c8eaf5ff6905eb9eeca5.

(cherry picked from commit bfeaa62dbc9d2a436e2d3413386ba38f48b5d502)

Resolves: RHEL-108587


dtardon (changes requested):
I don't think we need to backport this. It's not a bugfix, just a simple refactoring of the code.

PR #359, opened 2025-08-28
(RHEL-108576) shell completion: add kernel-identify/inspect verbs for bootctl

jamacku:
Follow-up for a05255981ba5b04f1cf54ea656fbce1dfd9c3a68 Follow-up for 3e0a3a0259324b4c40a9a62c8506fe683cd0273b

(cherry picked from commit 6a6d4c3f3c123a1cbb6770f1cae8c130a48333e1)

Resolves: RHEL-108576


dtardon (approved):
`bootctl` is probably only marginally interesting for RHEL users, hence I'm not sure this is really needed. OTOH, it's simple and unlikely to break anything (anything important in any case), so why not...

LGTM

PR #365, opened 2025-09-02
(RHEL-111628) Fix for "Unexpected error response from GetNameOwner(): Connection terminated"

brozs:
Resolves: RHEL-111628

### DIFF
--- a/src/core/dbus.c
+++ b/src/core/dbus.c
@@ -834,6 +834,8 @@ int bus_init_api(Manager *m) {
         if (r < 0)
                 return log_error_errno(r, "Failed to set up API bus: %m");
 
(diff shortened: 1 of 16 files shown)


mrc0mmand:
> RPM build failures do not appear to be related to this PR. The builders are looking for `python3dist(zstd)` (`python3-zstd`) RPM that is not available.

Yeah, the Packit fails are unrelated (that's an unfortunate side effect of using the latest systemd spec file for all 9.x releases; I'll try to work around this in the future).

The CentOS CI fail can be fixed by backporting https://github.com/redhat-plumbers/systemd-rhel9/commit/5ce6e9778a1f9708b4e52d143a0ed044a2a9e815.

PR #366, opened 2025-09-02
fstab-generator: fix options in systemd.mount-extra= arg

jamacku:
Fixes a bug introduced by 55365b0a233ae3024411fd0815ad930e20f6a3d6 (v254).

The arguments `(rd.)systemd.mount-extra` take a value that looks like `WHAT:WHERE[:FSTYPE[:OPTIONS]]`. The `OPTIONS` were parsed into a nulstr where a comma-separated c-string was expected. This leads to a bug where only the first option was taken into account by the generator.

For example, if you passed `systemd.mount-extra=/x:/y:baz:ro,defaults` to the kernel, `systemd-fstab-generator` would translate that into a nulstr: `ro\0defaults\0`.
Since methods processing options in the generator expected a comma-separated c-string, they would only see the first option, `ro` in this case.

(cherry picked from commit 06fadc4286fee6a7505a88659e5ae2e6f3ee60ba)

Resolves: RHEL-108580


dtardon (approved):
LGTM

PR #368, opened 2025-09-02
test-time-util: do more suppression of time zone checks

mrc0mmand:
The issue is directly triggered by tzdata-2024b, where the setting of timezone
started to fail and the tests stopped passing. But those timestamps in 1/1/1970
appear to have some problems already before:

  $ sudo date -s 'Thu 1970-01-01 13:00:01 WET'
  Thu Jan  1 03:00:01 PM EET 1970
  $ sudo date -s 'Thu 1970-01-01 12:00:01 WET'
  date: cannot set date: Invalid argument
  Thu Jan  1 02:00:01 PM EET 1970
  $ rpm -q tzdata
  tzdata-2024a-9.fc41.noarch

The same issue appears with other timezones. So move the first timestamp one
day forward to avoid the issue.

After the previous problem is solved, we also get the problem already seen
previously where the roundtrip returns a time that is off by one hour:

@86401000000 → Fri 1970-01-02 00:00:01 WET → @82801000000 → Thu 1970-01-01 23:00:01 WET
Assertion 'x / USEC_PER_SEC == y / USEC_PER_SEC' failed at src/test/test-time-util.c:415, function test_format_timestamp_impl(). Aborting.

Extend the override to suppress this.

(cherry picked from commit 3cf362f6f57b7d0b5f6b86a49316303b0dda7599)


mrc0mmand:
Oh well, the mkosi check is now green but CentOS CI fails due to desynchronized mirrors. I'll retry the failing CIs tomorrow, hopefully the mirrors sync by then.

mrc0mmand:
> Basically LGTM, but maybe we could drop some of the commits...

Thanks for the review! I dropped the mentioned commits, since they're indeed not required.

jamacku (approved):

PR #375, opened 2025-09-30
(RHEL-118215) timer: don't run service immediately after restart of a timer

mrc0mmand:
When a timer is restarted, don't reset the last_trigger field.
This prevents the timer from triggering immediately.

Fixes: #31231
(cherry picked from commit 3fc44a0f68412b649e16f12ff2f97a36c615457d)

Resolves: RHEL-118215


dtardon (approved):
I assume the last commit isn't really needed, but it's not a big deal...

LGTM

PR #409, opened 2026-03-26
(RHEL-166185) test: make test-fd-util more lenient when using fd_move_above_stdio()

mrc0mmand:
On s390x this test fails when the SUT uses the z90crypt kernel module,
as it's an another FD the test doesn't account for:

/* test_rearrange_stdio */
Successfully forked off 'rearrange' as PID 57293.
test_rearrange_stdio: r=0
/proc/57293/fd:
total 0
lrwx------. 1 root root 64 Apr  5 06:18 0 -> /dev/pts/0
lrwx------. 1 root root 64 Apr  5 06:18 1 -> /dev/pts/0
lrwx------. 1 root root 64 Apr  5 06:18 2 -> /dev/pts/0
lrwx------. 1 root root 64 Apr  5 06:18 3 -> /dev/z90crypt
rearrange terminated by signal ABRT.

Debugging this was pain, since the child process didn't log anything
once we closed stdout/stderr (for obvious reasons). Let's fix both
issues by switching logging to kmsg once we close stdin/stdout/stderr,
and also by making the test work fine when there are some extra FDs in
the child's environment.

(cherry picked from commit https://github.com/redhat-plumbers/systemd-rhel9/commit/a9805f8ca9c1561e373355fe7175579b31e1c08c)

---

This was merged to RHEL 9.8 in https://github.com/redhat-plumbers/systemd-rhel9/pull/373, but it fails on older releases as well, which is annoying. However, it's not annoying enough to deserve its own minor release, so let's pull this in with the next batch of fixes (if any) for RHEL 9.7.

### DIFF
--- a/src/test/test-fd-util.c
+++ b/src/test/test-fd-util.c
@@ -127,6 +127,7 @@ TEST(rearrange_stdio) {
 
         if (r == 0) {
                 _cleanup_free_ char *path = NULL;
+                int pipe_read_fd, pair[2];
                 char buffer[10];
 
                 /* Child */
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

PR #412, opened 2026-04-09
(RHEL-163870) Add some hardening to config file parser in nspawn (9.8)

dtardon:
Resolves: RHEL-163870


dtardon:
`rpm-build:centos-stream-10-s390x` appears to have problems, as many unit tests--which couldn't possibly be affected by this change--have failed with `SIGABRT`.

jamacku (approved):
LGTM

PR #419, opened 2026-04-16
(RHEL-137251) Add new `RestartMode=` option

dtardon:
Resolves: RHEL-137251


dtardon:
`rpm-build:centos-stream-9-s390x` is busted.

jamacku (approved):
LGTM

PR #421, opened 2026-04-22
(RHEL-164539) Increment start limit counter only when we can start the unit

dtardon:
Resolves: RHEL-164539


dtardon:
`centos-stream-9-s390x` is broken.

mrc0mmand (approved):
LGTM.

PR #422, opened 2026-04-27
(RHEL-127425) core: cleanup unit's dropin directories from global cache

msekletar:
When user creates dropin files via API (e.g. systemctl set-property ...) we put the dropin directory path into unit_path_cache. Drop those directories from the cache in unit_free() and prevent memory leak.

Follow-up for fce94c5c563b8f6ede2b8f7f283d2d2faff4e062.

Resolves: RHEL-127425

### DIFF
--- a/src/core/unit.c
+++ b/src/core/unit.c
@@ -603,6 +603,8 @@ static void unit_remove_transient(Unit *u) {
         if (!u->transient)
                 return;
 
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

dtardon:
`rpm-build:centos-stream-9-s390x` is busted.

PR #424, opened 2026-04-28
(RHEL-166186) fstab-generator: support swap on network block devices

mrc0mmand:
Teach swap units to support the _netdev option as well, which should make swaps on iSCSI possible. This mirrors the logic we already have for regular mounts in both the fstab-generator and the core (mount.c/swap.c).

Co-developed-by: Claude Opus 4.6 <noreply@anthropic.com>
(cherry picked from commit 3d5bd67a2259e7a4edc27476d4cae049653c4414)

Resolves: RHEL-166186


dtardon (approved):
LGTM

dtardon:
`rpm-build:centos-stream-9-s390x` is broken.

PR #426, opened 2026-04-29
(RHEL-3631) udev: allow persistent storage rules for rbd devices

dtardon:
The RADOS Block Device (rbd) can be used as any other block device with further layers on top of it, hence allow the common persistent storage rules to apply, including watching for changes.

(cherry picked from commit cbe65d38cf0a2e55cdba75871de108bc505a7095)

Resolves: RHEL-3631


dtardon:
`rpm-build:centos-stream-9-s390x` is broken.

mrc0mmand (approved):
LGTM

PR #427, opened 2026-05-04
(RHEL-25518) udev/net_id: introduce naming scheme for RHEL-9.9

jamacku:
rhel-only: policy

Resolves: RHEL-25518


mrc0mmand (approved):
LGTM.

PR #428, opened 2026-05-04
(RHEL-159165) core: validate input cgroup path more prudently

tcornell-bus:
(cherry picked from commit efa6ba2ab625aaa160ac435a09e6482fc63bdbe8)

Resolves: RHEL-159165

### DIFF
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -666,6 +666,12 @@ static int method_get_unit_by_control_group(sd_bus_message *message, void *userd
         if (r < 0)
                 return r;
 
+        if (!path_is_absolute(cgroup))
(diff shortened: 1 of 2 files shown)


mrc0mmand:
(There's currently an outage across multiple GH services (https://www.githubstatus.com/) including Actions, so ignore the GHA fails for now, I'll restart the jobs once this is over.)

mrc0mmand:
The CentOS CI fail is a legitimate, but unrelated test issue (we, unfortunately, test everything on the latest C9S here, so mdadm here has a slightly different behaviour than it would have on 9.6). The fix for this thankfully simple and backwards-compatible, just cherry-pick systemd/systemd@16406420ea449b75e70a7dced05d7b98bc0f5376 (as was done for later releases in https://github.com/redhat-plumbers/systemd-rhel9/pull/340).

mrc0mmand (approved):
LGTM, thanks!

PR #435, opened 2026-07-03
(RHEL-180934) [follow-up to] - udev: check for invalid chars in various fields received from the kernel

jamacku:


dtardon (approved):
Three commits to remove one `if` statement 😮

LGTM anyway

PR #437, opened 2026-07-09
(RHEL-193714) nss-systemd: avoid ELF TLS for recursion guard

msekletar:
libnss_systemd currently uses a thread_local recursion guard to avoid re-entering nss-systemd during NSS lookups.
Since libnss_systemd.so.2 is loaded lazily by glibc, accessing ELF TLS may trigger dynamic TLS allocation in __tls_get_addr(). Under allocation failure conditions, glibc terminates the process from the dynamic loader instead of allowing the NSS module to return a normal failure. Replace the recursion guard with POSIX thread-specific data to preserve the same per-thread semantics while avoiding ELF TLS in the NSS module. Note that pthread_setspecific() may still allocate internally on first use per thread. The key improvement is that any such failure is returned as a normal error code rather than terminating the process from inside the dynamic loader.

Related: #42559
(cherry picked from commit 19bd80e29a02b4f8c9543370eb4a16c014d497f3)

Resolves: RHEL-193714

### DIFF
--- a/src/nss-systemd/nss-systemd.c
+++ b/src/nss-systemd/nss-systemd.c
@@ -1057,28 +1057,72 @@ enum nss_status _nss_systemd_initgroups_dyn(
         return any ? NSS_STATUS_SUCCESS : NSS_STATUS_NOTFOUND;
 }
 
-static thread_local unsigned _blocked = 0;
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

PR #441, opened 2026-07-17
(RHEL-208860) resolved: replace assert() with error return in DNSSEC verify functions

### LINKED ISSUE #41569: resolved: replace assert() with error return in DNSSEC signature verification
`dnssec_rsa_verify_raw()` at line 103 and `dnssec_ecdsa_verify_raw()` at line 235 of `src/resolve/resolved-dns-dnssec.c` use `assert()` to check conditions derived from DNS record content.

Line 103:
```c
assert((size_t) RSA_size(rpubkey) == signature_size);
```
`RSA_size(rpubkey)` is determined by the DNSKEY modulus, `signature_size` by the RRSIG record. These are two independent records. `dnssec_rrsig_match_dnskey()` matches on algorithm and key_tag but does not check key size vs signature size consistency, so a malformed pair could get through.

Line 235:
```c
assert(EC_KEY_check_key(eckey) == 1);
```
On OpenSSL < 3.0, `EC_POINT_oct2point()` can succeed for coordinates not on the curve, leaving the validation to `EC_KEY_check_key()`. On 3.0+ this path is dead since `oct2point` already rejects such points.

Neither is easy to trigger because the DNSKEY must pass DS chain validation first. Still, both are `assert()` on data that ultimately originates from the network. Replacing them with `if (...) return -EINVAL` would be cleaner. `EVP_PKEY_verify()` and `ECDSA_do_verify()` handle the mismatch fine on their own, making both asserts redundant anyway.

dtardon:
dnssec_rsa_verify_raw() asserts that RSA_size(key) matches the RRSIG signature size, and dnssec_ecdsa_verify_raw() asserts that EC_KEY_check_key() succeeds. Both conditions depend on parsed DNS record content. Replace with proper error returns.

The actual crypto verify calls (EVP_PKEY_verify / ECDSA_do_verify) handle mismatches fine on their own, so the asserts were also redundant.

While at it, fix the misleading "EC_POINT_bn2point failed" log message that actually refers to an EC_KEY_set_public_key() failure.

Fixes: https://github.com/systemd/systemd/issues/41569 (cherry picked from commit dd80e5a348bdb8185e040f66ede00fd4ffdee777)

Resolves: RHEL-208860

### DIFF
--- a/src/resolve/resolved-dns-dnssec.c
+++ b/src/resolve/resolved-dns-dnssec.c
@@ -125,7 +125,8 @@ static int dnssec_rsa_verify_raw(
                 return -EIO;
         e = m = NULL;
 
-        assert((size_t) RSA_size(rpubkey) == signature_size);
+        if ((size_t) RSA_size(rpubkey) != signature_size)
+                return -EINVAL;
 
         epubkey = EVP_PKEY_new();
         if (!epubkey)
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM

PR #442, opened 2026-07-20
(RHEL-155457) ci: bump the mkosi job to Ubuntu Noble

mrc0mmand:
This should help with the missing pacman-package-manager package in the mkosi job.

rhel-only: ci
Related: RHEL-155457


jamacku (approved):
Thank you, LGTM

PR #443, opened 2026-07-23
(RHEL-213655) core: make manager event loop rate limit configurable

siteshwar:
Co-developed-by: Claude Opus 4.6 <noreply@anthropic.com>
(cherry picked from commit e1c63fd4c0657a7fc7b749c64283ab4dbc6fb39e)

Resolves: RHEL-213655

### DIFF
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -524,6 +524,10 @@ node /org/freedesktop/systemd1 {
       @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
       readonly t DefaultTasksMax = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
+      readonly t EventLoopRateLimitIntervalUSec = ...;
+      @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
+      readonly u EventLoopRateLimitBurst = ...;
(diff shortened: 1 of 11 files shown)


man/org.freedesktop.systemd1.xml
```diff
@@ -10877,4 +10889,11 @@ $ gdbus introspect --system --dest org.freedesktop.systemd1 \
   </refsect1>
 
   <xi:include href="org.freedesktop.locale1.xml" xpointer="versioning"/>
+  <refsect1>
```
siteshwar: I am assuming this can be removed.

src/core/manager-serialize.c
dtardon, line 556: The change in handling deserialization of `dump-ratelimit` (using the newly added `deserialize_ratelimit()`) is missing.

It's not a big deal, as the old code still works, but it means duplication.

dtardon (changes requested):

man/org.freedesktop.systemd1.xml
mrc0mmand: Yes, I don't think we want this in RHEL 9.

src/core/manager-serialize.c
dtardon: It's better now, but please add the chunk to the _first_ commit (where it belongs).

dtardon (changes requested):

dtardon (approved):
LGTM

PR #450, opened 2026-08-11
(RHEL-155457) test: include util.sh needed for assert_*() functions 

mrc0mmand:
This is needed by the test backported by
https://github.com/redhat-plumbers/systemd-rhel9/commit/219f89250a9d41c357eb1224f624ae3cf08cf4e9.

In upstream the include was introduced as part of
https://github.com/redhat-plumbers/systemd-rhel9/commit/5db456d06511ed537e5f6870e9d6b1afe50587c8, but we don't have the
necessary prerequisites in RHEL 9, so let's add just the include
manually.

rhel-only: ci
Related: RHEL-155457

---

Also, drop the forgotten `set +e` thanks to which we didn't notice the issue earlier.


jamacku (approved):
LGTM

PR #451, opened 2026-08-18
(RHEL-33815) udev-node: optimize device node symlink creation

siteshwar:
If multiple devices requested the same device node symlink with the same
priority, then previously we read O(N^2) of files saved in
/run/udev/links.

This makes if the requested symlink already exists with equal or higher
priority, then the symlink is kept, and skip to read all existing files,
except for one related to the current device node, in /run/udev/links.
Hence, the total amount of file read becomes O(N).

This improves performance of testcase_simultaneous_events_2 added by the
previous commit about 30%.
Before (32.8 sec):
```
 ## 3 iterations start: 11:13:44.690953163
 ## 3 iterations end: 11:14:17.493974927
```
After (23.8 sec):
```
 ## 3 iterations start: 11:17:53.869938387
 ## 3 iterations end: 11:18:17.624268345
```

This is based on the idea and analysis by Franck Bui.

Replaces #25839.

Co-authored-by: Franck Bui <fbui@suse.com>
(cherry picked from commit 331aa7aa15ee5dd12b369b276f575d521435eb52)

Resolves: RHEL-33815

### DIFF
--- a/src/basic/locale-util.c
+++ b/src/basic/locale-util.c
@@ -17,6 +17,7 @@
 #include "fileio.h"
 #include "hashmap.h"
 #include "locale-util.h"
+#include "missing_syscall.h"
 #include "path-util.h"
 #include "set.h"
 #include "string-table.h"
@@ -294,6 +295,12 @@ bool is_locale_utf8(void) {
 
         if (cached_answer >= 0)
                 goto out;
+
+        /* This function may be called from libsystemd, and setlocale() is not thread safe. Assuming yes. */
+        if (gettid() != raw_getpid()) {
+                cached_answer = true;
+                goto out;
+        }
 
         if (!setlocale(LC_ALL, "")) {
                 cached_answer = true;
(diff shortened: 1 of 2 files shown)


siteshwar:
Manually verified the fix works on `c9s` through the [reproducer](https://redhat.atlassian.net/browse/RHEL-33815?focusedCommentId=17692447):

Before the fix:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100.00    0.015343           7      2015         6 readlinkat
------ ----------- ----------- --------- --------- ----------------
100.00    0.015343           7      2015         6 total
```

After the fix:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100.00    0.000593          17        34         6 readlinkat
------ ----------- ----------- --------- --------- ----------------
100.00    0.000593          17        34         6 total
```

siteshwar:
The CI failures are unrelated to the changes in this pull request and mentions of upstream commits in this pull request are not needed for RHEL 9. I am marking this pull request ready for review as I have verified the changes manually.

PR #452, opened 2026-08-19
(RHEL-212086) test: rename the start limit subtest

mrc0mmand:
It uses the current upstream naming, unfortunately in RHEL 9 we still use the previous one, so the test was not picked up by the TEST-07-PID1's main test script.

rhel-only: ci
Related: RHEL-212086

Follow-up for f18478c89e0f88eab88bb7f79ac021b6ebd28cb4.


jamacku (approved):

dtardon:
Oops...

dtardon (approved):
LGTM

PR #453, opened 2026-08-19
test: rename the start limit subtest

mrc0mmand:
It uses the current upstream naming, unfortunately in RHEL 9 we still
use the previous one, so the test was not picked up by the
TEST-07-PID1's main test script.

Also, unmask & enable the systemd-resolved service, as the old testing
framework masks it by default.

rhel-only: ci
Related: RHEL-164539

Follow-up for 4060cdad388b0ae658f2024633b842a46c37962e.


jamacku (approved):
LGTM

